Issue #202, opened 2023-12-01
vless


greenhat616:
Might we should impl vless first?

vless basic stream is based on vmess, ref:https://github.com/MetaCubeX/mihomo/blob/Alpha/adapter/outbound/vless.go
and grpc based on gun. I can draft a `gun` transport first

ibigbug:
i think we already have some level of gRPC support https://github.com/Watfaq/clash-rs/blob/c84d26251b9bb2e648a97f9e15accce903884b6c/clash_lib/src/proxy/transport/grpc.rs#L49

i had a quick look at `gun`, which seems to be something similar(same) to the grpc transport?

ibigbug:
I'm setting up a grpc proxy to test it out. I don't think that code has been tested ever.

ibigbug:
it looks `gun` is grpc over h2.

i'm working on the PR #203 however there seems to be something wrong. i'll try to fix it later.

ibigbug:
I'm kind of stuck - https://github.com/Watfaq/clash-rs/pull/203

if you are more familiar maybe you can take a look, or I'll get back when I have more time to dig in

greenhat616:
> I'm kind of stuck - #203
> 
> if you are more familiar maybe you can take a look, or I'll get back when I have more time to dig in

Okay. I can have a try tomorrow

ibigbug:
the gun/gRPC is done. maybe we should promote this - however is this in general more popular/better then vmess?

captainnx:
usually vless is much batter than vmess

greenhat616:
> the gun/gRPC is done. maybe we should promote this - however is this in general more popular/better then vmess?

It seems that Vmess, SS, vless, hysteria are widely used by Proxies Provider.
So it highly  recommended to  implement it in advanced priority.

What's more, SS 2022 could be easy  to introduced due to the SS library we are directly imported.

yokiyokiyoki:
vless first！

charley008:
看来还是没有vless reality

ibigbug:
#903

PR #203, opened 2023-12-02
implement grpc for trojan and vmess

ibigbug:
#69 #202


ibigbug:
it wouldn't send more data after the init request is done for some reason

image

greenhat616:
> Yeah I think there’s something wrong in how they handle the array boundry.
> 
> Another thing can be done is maybe to try the grpc with Trojan (I’ve put it in too in this PR) and see if it’s actually the problem of our grpc impl or only grpc with vmesd is the issue

I test currect pr in trojan with gRPC( xray is the backend), using the config:
```
{
  "log": {
    "loglevel": "debug" // Level of log
  },
  "inbounds": [
    {
      "port": 12312,
      "protocol": "trojan",
      "settings": {
        "clients": [
          {
            "password": "password",
            "email": "love@xray.com",
            "level": 0
          }
        ],
        "fallbacks": [
          {
            "dest": 80
          }
        ]
      },
      "streamSettings": {
        "network": "grpc",
        "grpcSettings": {
          "serviceName": "abc"
        },
        "security": "none"
      }
    }
  ],
  "outbounds": [
    {
      "protocol": "freedom",
      "settings": {}
    }
  ]
}
```
It seems small page, such as `www.baidu.com` can get a response immediately. and huge page like `www.aliyun.com`, `www.google.com`, or a pure huge http page `http://www.galgamez.me/bbs/`, could get a buffer out of range error:
```
split_to out of bounds: 4096 <= 4054
  2023-12-05T14:58:06.236309Z ERROR clash_lib: panic hook: PanicInfo { payload: Any { .. }, message: Some(split_to out of bounds: 4096 <= 4054), location: Location { file: "C:\\Users\\a632079\\.cargo\\registry\\src\\rsproxy.cn-0dccff568467c15b\\bytes-1.5.0\\src\\bytes_mut.r
```
image

ibigbug:
Ok nice we are making progresses. 👍

I'll take a look later. Thanks!

ibigbug:
@greenhat616 i think it's working now. feel free to give it a try on master branch

PR #206, opened 2023-12-06
simplify the method of stream forward  in `dispatch_stream`

xmh0511:


clash_lib/src/common/io.rs
```diff
@@ -1,291 +1,383 @@
...
+        let mut buf = Vec::new();
+        buf.resize(size, 0);
         loop {
-            match a_to_b {
-                TransferState::Running(buf) => {
-                    let res = buf.poll_copy(cx, a.as_mut(), b.as_mut());
-                    match res {
-                        Poll::Ready(Ok(count)) => {
-                            *a_to_b = TransferState::ShuttingDown(count);
+            match tokio::time::timeout(a_to_b_timeout_duration, a_reader.read(&mut buf[..])).await {
```
ibigbug: have you tested this?

does this not break the long running connections ?

xmh0511: As exposed by the signature `copy_buf_bidirectional_with_timeout`, which is specified by `timeout`, and this function is only called in `dispatch_stream` in https://github.com/Watfaq/clash-rs/blob/master/clash_lib/src/app/dispatcher/dispatcher.rs#L138, which specifies the `Duration::from_secs(10)`, so this is not for long-running connection. The current implementation does work for me.

ibigbug: i think you are confused with the timeout. it's mean to be the connection idle timeout.

ibigbug: can you test if your change can successfully request `https://httpbin.yba.dev/drip\?duration\=11` ?

xmh0511: I am not confused by `timeout`, [tokio::time::timeout](https://teaclave.apache.org/api-docs/crates-app/tokio/time/fn.timeout.html) does mean the read should read data within the specified time duration. In other words, if the read future is idle over a specified time, the read future is cancelled.

ibigbug: here we need the idle timeout, not to timeout for a long entire io duration.

xmh0511: > here we need the idle timeout, not to timeout for a long entire io duration.

[tokio::time::timeout](https://teaclave.apache.org/api-docs/crates-app/tokio/time/fn.timeout.html) exactly is the idle timeout.

ibigbug: can you test if your change can successfully request https://httpbin.yba.dev/drip\?duration\=11 ?

xmh0511: > can you test if your change can successfully request [https://httpbin.yba.dev/drip\?duration\=11](https://httpbin.yba.dev/drip%5C?duration%5C=11) ?

I tried to use the real `config.yaml` to start the program, which gives the error 
````
ERROR clash_lib: panic hook: PanicInfo { payload: Any { .. }, message: Some(called `Result::unwrap()` on an `Err` value: InvalidConfig("unsupported network: grpc")), location: Location { file: "clash/src/main.rs", line: 68, col: 6 }, can_unwind: true, force_no_backtrace: false }
````

But, this ` InvalidConfig("unsupported network: grpc"))` is not caused by this change.

Comment (ibigbug) [reply clash_lib/src/common/io.rs]: try start with https://github.com/Watfaq/clash-rs/blob/master/clash/tests/data/config/simple.yaml

Comment (xmh0511) [reply clash_lib/src/common/io.rs]: > try start with https://github.com/Watfaq/clash-rs/blob/master/clash/tests/data/config/simple.yaml

I have successfully started the program with this `simple.yaml`,
```
  2023-12-06T12:05:41.916730Z  INFO clash_lib::app::inbound::network_listener: HTTP TCP listening at: 0.0.0.0:8080
    at clash_lib/src/app/inbound/network_listener.rs:118

  2023-12-06T12:05:41.922244Z  INFO clash_lib::app::api: Starting API server at 127.0.0.1:9090
    at clash_lib/src/app/api/mod.rs:59

  2023-12-06T12:05:41.937131Z  INFO clash_lib: receive shutdown signal
    at clash_lib/src/lib.rs:275
```

then what do I need to do for the next step?

Comment (ibigbug) [reply clash_lib/src/common/io.rs]: so you never tested your change?

you just use the proxy to request the above url, something like

```
curl -x http://127.0.0.1:8080  https://httpbin.yba.dev/drip\?duration\=11
```

Comment (xmh0511) [reply clash_lib/src/common/io.rs]: @ibigbug As shown in the snapshot, this is ok.



For `https` website access, this is not supported by this library yet, however, this is not caused by this change.

Comment (ibigbug) [reply clash_lib/src/common/io.rs]: i'm talking about this url `https://httpbin.yba.dev/drip\?duration\=11`

Comment (xmh0511) [reply clash_lib/src/common/io.rs]: @ibigbug Response `curl -x http://127.0.0.1:8080 https://httpbin.yba.dev/drip\?duration\=11
**********%  `

Comment (xmh0511) [reply clash_lib/src/common/io.rs]: It can also respond curl: (56) Failure when receiving data from the peer, I suppose this is because of `idle timeout`? We may set the Duration longer than `10`.

Comment (ibigbug) [reply clash_lib/src/common/io.rs]: right, 11 might be too closed to the timeout.

if you can try with duration = 15 or some longer value you should see some error like this

```
curl: (56) Failure when receiving data from the peer
```

as it closes the conn premaruely

```
-> % curl -v -k -x socks5h://127.0.0.1:8889 https://httpbin.yba.dev/drip\?duration\=13
*   Trying 127.0.0.1:8889...
* Connected to 127.0.0.1 (127.0.0.1) port 8889 (#0)
* SOCKS5 connect to httpbin.yba.dev:443 (remotely resolved)
* SOCKS5 request granted.
* Connected to 127.0.0.1 (127.0.0.1) port 8889 (#0)
* ALPN: offers h2,http/1.1
* (304) (OUT), TLS handshake, Client hello (1):
* (304) (IN), TLS handshake, Server hello (2):
* (304) (IN), TLS handshake, Unknown (8):
* (304) (IN), TLS handshake, Certificate (11):
* (304) (IN), TLS handshake, CERT verify (15):
* (304) (IN), TLS handshake, Finished (20):
* (304) (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / AEAD-CHACHA20-POLY1305-SHA256
* ALPN: server accepted h2
* Server certificate:
*  subject: C=US; ST=California; L=San Francisco; O=Cloudflare, Inc.; CN=sni.cloudflaressl.com
*  start date: Feb 10 00:00:00 2023 GMT
*  expire date: Feb  9 23:59:59 2024 GMT
*  issuer: C=US; O=Cloudflare, Inc.; CN=Cloudflare Inc ECC CA-3
*  SSL certificate verify ok.
* using HTTP/2
* h2 [:method: GET]
* h2 [:scheme: https]
* h2 [:authority: httpbin.yba.dev]
* h2 [:path: /drip?duration=13]
* h2 [user-agent: curl/8.1.2]
* h2 [accept: */*]
* Using Stream ID: 1 (easy handle 0x14480bc00)
> GET /drip?duration=13 HTTP/2
> Host: httpbin.yba.dev
> User-Agent: curl/8.1.2
> Accept: */*
>
* Closing connection 0
curl: (56) Failure when receiving data from the peer

```

ibigbug: it's not about the Duration value, you are changing the timeout behaviour.

I think this PR causes error. I'll not further proceed this.

xmh0511: > I think this PR causes error. I'll not further proceed this.

This PR does not cause the error. I test it with the result you have said above. In this PR, the timeout is set to be `10`, and with the duration `11`, I see the result
> curl: (56) Failure when receiving data from the peer

Then I change the timeout with `15` in the source code, and I see the result `*******`.

ibigbug: that's exactly the point.

sorry I got something else to fix, I'll not reply to this thread.

xmh0511: What the "connection idle-timeout" do you mean here? My understanding is if the connection cannot read data within `X` time, the connection is considered to be idle, the `timeout` just specifies what maximum time the connection can be in this idle state before reading.

Incidentally, in your original implementation, whatever the number is given in `https://httpbin.yba.dev/drip\?duration\=number`, it always responds `********` even though the Duration in `copy_buf_bidirectional_with_timeout` is set to be `10`.


Cite the meaning of [idle-timeout](https://www.watchguard.com/help/docs/help-center/en-US/Content/en-US/WG-Cloud/Devices/managed/firewall_policies_config_idle_timeout.html) here
> Idle timeout is the maximum length of time that a TCP connection can stay active **when no traffic is sent through the connection**.

ibigbug: add a numbytes param to the url you'll see more *s https://httpbin.yba.dev/drip\?duration\=20\&numbytes\=20

ibigbug: ok i finished the other stuff. now i'm getting back to this.

as you mentioned in the link, this timeout needs to be the duration during which *no traffic is sent*, not the entire time it takes to finish the connection.

think about you need to download a large file, it take 2hrs, but in the 2hrs, there's always data being transfered, you can't just give it a 10s timeout or whatever.

xmh0511: > think about you need to download a large file, it take 2hrs, but in the 2hrs, there's always data being transfered, you can't just give it a 10s timeout or whatever.

Consider this case, the peer sends a byte every 1hrs or 12hrs, is this permitted? In this implementation, the timeout just regulates that the peer must send some traffic within every specified time. In other words, if the peer occupies the connection and does not send some traffic within the specified time, the connection will be closed. Every send traffic within the specified time will make the timeout reset(i.e. recalculate the time)

I think the meaning of the timeout is to prevent "Slow HTTP Attack"/"Slowloris". If I didn't understand correctly, please tell me what the necessary logic here, and I will try to implement it in this PR.

xmh0511: After reviewing the original implementation carefully, the timeout logic is, that only if one direction is complete, it will set the timeout for the other direction. In other words, if two directions are both pending(i.e. no traffic can be accepted), there is no timeout imposed on them. Is this really what we want here?

xmh0511: To clarify what the `#1` implementation does in the PR, I need to say the implementation cannot close a long-running traffic. 

> think about you need to download a large file, it take 2hrs, but in the 2hrs, there's always data being transfered, you can't just give it a 10s timeout or whatever.

The specified timeout `10` does not mean the whole data communication must be completed within `10s`, it's a misunderstanding of the implementation. The `#1` implementation means when we do `read` in each loop, some data traffic must be read within `10s`, and in the next loop, the timeout is re-calculated for the next read. 

Consider this scenario, after `A` and `B` have been connected, `B` writes some bytes to `A` but is not end, then `B` does not send any traffic to `A` within 2hrs, and after 2hrs, `B` then continues to send data to `A`. Isn't the duration of the 2hrs considered an idle connection?

xmh0511:
@ibigbug I have `1:1` implemented the original logic of forwarding stream in a concise way, could you take a look at the latest implementation and make further progress?

ibigbug:
what are you exactly trying to do in your patch? could you add a description, what's the issue and why your change is useful?

xmh0511:
> what are you exactly trying to do in your patch? could you add a description, what's the issue and why your change is useful?

I'm trying to implement the original logic concisely and effectively. The original logic is a hand-write state machine, which mixes the `a_to_b` and `b_to_a` into a single machine state, which results in whenever one task is wakened up, the other task will be polled again, especially when both of them are in `TransferState::Running` state, and in this state, the other hand `poll_read` will be called again even though it previously returns `Poll::Pending` and the current call is not triggered by the `waken` of it, which is not a correct use of `poll_read`. 

In this patch, I divorce `a_to_b` and `b_to_a` into separate tasks, and they do not influence each other when one of them is wakened, which is effective and maintainable.

xmh0511:
> what's the difference of ready! two futures? isn't it the same thing.

When a future is wakened, the other future will be polled again even though it is not notified by its `ctx.wake()`. Specifically, when the `poll_read` of  `a_to_b` is wakened by its `ctx.wake()`, that means it has something to be read, however, this waken will make `poll_read` of `b_to_a` be invoked again and it can still return `Poll::Pending`.

In this patch, the implementation can make sure that they are wakened again by their `ctx.wake()`, respectively. Furthermore, the patch eliminate the complex hand-written machine state in the source code.

ibigbug:
i think this line is problematic - the await is block, and in the case the 2nd future is errored/finished before the 1st, it'd never know https://github.com/Watfaq/clash-rs/pull/206/files#diff-85611fef80536aedcb4665af1a4b566df2341c0e3b6cdf60b28eaf9a1d81ba92R433

ibigbug:
> When a future is wakened, the other future will be polled again even though it is not notified by its ctx.wake(). 

i think this is not a big issue https://rust-lang.github.io/async-book/02_execution/02_future.html

it's how Rust tell people to run 2 futures (the Join example)

xmh0511:
I think the main goal of this patch is the following:
> 1. Implement the forwarding stream in a briefer way, the original impl needs 280 lines while this patch just 117 lines
> 2. Eliminate the complex hand-written state machine to make the implementation more readable and maintainable
> 3. This patch may have an effective performance since it does not have meaningless awaken, `a_to_b` and `b_to_a` are separately run on their own task space.

Issue #191, opened 2023-11-24
Missing Web api


greenhat616:
Original API abstract definition:
```go
r.Get("/", hello)
r.Get("/logs", getLogs)
r.Get("/traffic", traffic)
r.Get("/memory", memory)
r.Get("/version", version)
r.Get("/configs", getConfigs)
r.Put("/configs", updateConfigs)
r.Post("/configs/geo", updateGeoDatabases)
r.Patch("/configs", patchConfigs)
r.Get("/proxies", getProxies)
r.Get("/proxies/{name}", getProxy)
r.Get("/proxies/{name}/delay", getProxyDelay)
r.Put("/proxies/{name}", updateProxy)
r.Get("/group", getGroups)
r.Get("/group/{name}", getGroup)
r.Get("/group/{name}/delay", getGroupDelay)
r.Get("/rules", getRules)
r.Get("/connections", getConnections)
r.Delete("/connections", closeAllConnections)
r.Delete("/connections/{id}", closeConnection)
r.Get("/providers/proxies", getProviders)
r.Get("/providers/proxies/{providerName}", getProvider)
r.Put("/providers/proxies/{providerName}", updateProvider)
r.Get("/providers/proxies/{providerName}/healthcheck", healthCheckProvider)
r.Get("/providers/proxies/{providerName}/{name}", getProxy)
r.Get("/providers/proxies/{providerName}/{name}/healthcheck", getProxyDelay)
r.Get("/providers/rules", getRuleProviders)
r.Put("/providers/rules/{name}", updateRuleProvider)
r.Post("/cache/fakeip/flush", flushFakeIPPool)
r.Get("/dns/query", queryDNS)
r.Post("/restart", restart)
r.Post("/upgrade", upgradeCore)
r.Post("/upgrade/ui", updateUI)
```

These api impl are missing:
* `Get /memory`
* `POST /configs/geo`
* `/group` group
* `/cache` group
* `/restart` group
* `/providers/rules` group
* `GET /dns/dns` should be rename to `/dns/query`
*  `/upgrade` group

ibigbug:
are some of these from the premium core? i don't remember seeing all of them from the open sourced clash snapshot

greenhat616:
> are some of these from the premium core? i don't remember seeing all of them from the open sourced clash snapshot
You reminded me that indeed Clash Core is missing these interfaces. they are premium and meta only.

ibigbug:
[...]

PR #220, opened 2023-12-25
add config reload api

ibigbug:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link
#191 

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -831,7 +831,6 @@ dependencies = [
  "shadowsocks",
  "smoltcp",
  "socket2 0.5.5",
- "state",
  "tempfile",
  "thiserror",
  "tokio",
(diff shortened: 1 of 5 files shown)


ibigbug:
@greenhat616

greenhat616:
I tested it with nyanpasu dev, and server returned a 202(Accepted) error.

I think it should be blocking until the config refreshed, just as the meta does?
https://github.com/MetaCubeX/mihomo/blob/Alpha/hub/route/configs.go

ibigbug:
is the config actually reloaded?

202 is not an error, and i think it shouldn't matter if this API is blocking/async. the other clash i think it returns 204 instead

greenhat616:
> is the config actually reloaded?
> 
> 202 is not an error, and i think it shouldn't matter if this API is blocking/async. the other clash i think it returns 204 instead

Yes, I know it is intentional. It was caught by frontend request lib, such as `axios` or `fetch`. I can process it manually in next patch of nyanpasu. But users counld not get switch profiles until updates get released, though we have a core updater.

Issue #246, opened 2024-01-01
add integration tests


ibigbug:

VendettaReborn:
How about test with docker? Similar to the [docker test](https://github.com/MetaCubeX/mihomo/blob/meta-main/test/docker_test.go) in clash.meta

I've implemented one in my local environment, it works well, and i've found a bug in http2 transport (shall be fixed in another issue)

ibigbug:
yes that's a good option. we also have some proxy server setups here https://github.com/Watfaq/clash-rs/tree/master/docker, which might be useful to run tests

VendettaReborn:
yep, i see that. But that need developers' manual operations, right? I'd like to run/destroy the docker container in the unit test code, so the correctness can be verified with `cargo test` .

ibigbug:
yeah it's definitely good to automate things as much as possible. however not sure whether they should be unit tests, maybe we should add separate integration tests https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html

VendettaReborn:
I shall create a PR for this feature

PR #324, opened 2024-03-17
feat: add docker test for current protocol(vmess, ss, trojan)

VendettaReborn:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link

https://github.com/Watfaq/clash-rs/issues/246

### 💡 Background and solution

- background: currently, there is no verification for the protocol implementations
- solution: migrate docker test logic in clash.meta to clash-rs

### To Be Discussed

- whether we need this test method
- should we add some docs or setup codes(to pull the images)
- whether we should run ignore the docker test in github actions


clash_lib/src/proxy/utils/test_utils/docker_runner.rs
```diff
@@ -0,0 +1,158 @@
...
+use bollard::Docker;
+
+use bollard::exec::{CreateExecOptions, StartExecResults};
+use bollard::image::CreateImageOptions;
+
+use anyhow::Result;
+use futures::{Future, StreamExt, TryStreamExt};
+
+pub struct DockerTestRunner {
+    _instance: Docker,
```
ibigbug: why `_`?

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
```diff
@@ -0,0 +1,158 @@
...
+    id: String,
+}
+
+impl DockerTestRunner {
+    pub async fn new<
+        'a,
+        T1: serde::ser::Serialize + Into<std::string::String> + std::fmt::Debug + Clone,
```
ibigbug: is this `T1` is essentially a `String`? https://docs.rs/bollard/latest/bollard/image/struct.CreateImageOptions.html 

can we simplify the type signature for this function, same to the `Z` below

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
```diff
@@ -0,0 +1,158 @@
...
+    {
+        host_config.network_mode = Some("host".to_owned());
+    }
+    host_config.port_bindings = Some(
+        [
+            (
+                "10002/tcp".to_owned(),
+                Some(vec![PortBinding {
+                    host_ip: Some("0.0.0.0".to_owned()),
+                    host_port: Some("10002".to_owned()),
```
ibigbug: would there be port binding conflict when the tests were run in parallel? https://doc.rust-lang.org/book/ch11-02-running-tests.html#running-tests-in-parallel-or-consecutively

ibigbug (commented):
thanks!

couple of questions

ibigbug:
we should also setup Docker on the CI runners?

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+pub fn mount_config(pairs: &[(&str, &str)]) -> Vec<Mount> {
+    pairs
+        .iter()
+        .map(|(src, dst)| Mount {
+            target: Some(dst.to_string()),
+            source: Some(src.to_string()),
+            typ: Some(bollard::secret::MountTypeEnum::BIND),
+            read_only: Some(false),
```
ibigbug: read_only: true?

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
VendettaReborn: Yep, I found that problem, too, after the push of this PR, since it utilize `#[tokio::test]`.

Will a atomic counter be helpful? So we can calculate the port at the beginning of each test. An there shall be no conflict.

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
VendettaReborn: Mostly, this type is String.

The type is written in accordance with the type in bollard, the docker crate.

And I think you are right, we can simplify the signature here.

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
VendettaReborn: Oh, there is a story. 
You see, but when I first create the DockerTestRunner, i add `DockerTestRunner::exec` to execute docker commands, it needs the docker instance, so the instance shall be kept in `DockerTestRunner`.
But then, i found that, actually, these docker images all provide an entrypoint, so it's ok to just pass the startup parameters and don't do the `docker exec` job. So i turn the `instance` into `_instance`. 

So shall i just delete the `_instance` & `DockerTestRunner::exec` to simplify the code?

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
VendettaReborn: got it

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
ibigbug: Yeah if the entrypoint/cmd does the thing already, we can just use them

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
ibigbug: That should work too. 

Or we can maintain a hardcoded port lists for each test somewhere in the code and assign them while testing different protocols. As long as they are different. 

Which ever works and is simple to implement

VendettaReborn: I prefer an auto-increment  & global atomic counter(as the port). It's easy to implement, too.

VendettaReborn: I use serial_test instead. To change the port in different test is a bad idea, since the proxy server's port is hardcoded in the config file.

clash_lib/Cargo.toml
```diff
@@ -27,7 +27,7 @@ lru_time_cache = "0.11"
 hyper = { version = "0.14.28", features = ["http1","http2","client", "server", "tcp"] }
 http = { version = "1.1" }
 httparse = "1.8.0"
-h2 = "0.4.2"
+h2 = "0.4.3"
```
ibigbug: are these changes picked from master branch? i don't think you should have these if you didn't modify them

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
VendettaReborn: fixed

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
VendettaReborn: fixed

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
VendettaReborn: fixed

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
```diff
@@ -0,0 +1,237 @@
+//! This example will run a non-interactive command inside the container using `docker exec`
```
ibigbug: update the comment maybe

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
```diff
@@ -0,0 +1,237 @@
...
+            .id;
+        docker.start_container::<String>(&id, None).await?;
+        Ok(Self {
+            instance: docker,
+            id,
+        })
+    }
+
+    #[allow(unused)]
+    pub async fn exec(&self, cmd: Vec<&str>) -> anyhow::Result<()> {
```
ibigbug: please remove these if not used

clash_lib/src/proxy/vmess/mod.rs
```diff
@@ -94,7 +94,7 @@ impl Handler {
                 };
 
                 let h2_builder = Http2Config {
-                    hosts: vec![self.opts.server.clone()],
+                    hosts: opt.host.clone(),
```
ibigbug: seems your merging master failed and brought in unexpected changes

clash_lib/src/proxy/utils/test_utils/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+                Err(e) => error!("Failed to to proxy: {}", e),
+            },
+            Err(e) => error!("Failed to accept connection: {}", e),
+        }
+        debug!("proxy task finished");
+    });
+
+    let _ = join!(proxy_task, target_local_server_handler);
+
+    Ok(())
```
ibigbug: would this ping_pong_test always return success result? is there a way for us to know when it's failed?

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
```diff
@@ -0,0 +1,237 @@
...
+        }
+    }
+
+    // will make sure the container is cleaned up after the future is finished
+    pub async fn run_and_cleanup(
+        self,
+        f: impl Future<Output = anyhow::Result<()>>,
+    ) -> anyhow::Result<()> {
+        let fut = Box::pin(f);
+        let res = fut.await;
```
ibigbug: maybe give this `fut.await` a time out just in case it blocks.

and cargo doesn't seem to kill timeout tasks either https://stackoverflow.com/questions/73131115/unit-test-timeout-in-rust#comment134752168_73133545

ibigbug (commented):
couple of comments

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
ibigbug: yeah we can pre process the config files too with dyn ports, but that adds extra work. serial should be fien for now until the entire tests runs too long

clash_lib/Cargo.toml
VendettaReborn: yeah, it was picked from master, changed by the dependency bot

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
VendettaReborn: get

clash_lib/src/proxy/utils/test_utils/docker_runner.rs
VendettaReborn: it has been removed

clash_lib/src/proxy/utils/test_utils/mod.rs
```diff
@@ -42,26 +47,30 @@ pub async fn ping_pong_test(handler: Arc<dyn OutboundHandler>, port: u16) -> any
...
 
     let target_local_server_handler = tokio::spawn(async move {
-        match listener.accept().await {
-            Ok((stream, _)) => destination_fn(stream).await,
-            Err(e) => {
-                // Handle error e, log it, or ignore it
-                error!("Failed to accept connection: {}", e);
-                Err(anyhow!("Failed to accept connection: {}", e))
-            }
+        loop {
```
ibigbug, line 69: @VendettaReborn this should be a loop

ibigbug: for our case it might be fine as we only have 1 conn.

ibigbug: it was actually failing for me on macos as the proxy server inside container can't connect to the listener outside the vm on the host mac machine

ibigbug:
the condition is failing on Windows

```
   2 |  CLASH_RS_CI=true cargo test --all --all-features
     |  ~~~~~~~~~~~~~~~~
     | The term 'CLASH_RS_CI=true' is not recognized as a name of a cmdlet, function, script file, or executable
     | program. Check the spelling of the name, or if a path was included, verify that the path is correct and try
     | again.
Error: Process completed with exit code 1.

```

VendettaReborn:
> the condition is failing on Windows
> 
> ```
>    2 |  CLASH_RS_CI=true cargo test --all --all-features
>      |  ~~~~~~~~~~~~~~~~
>      | The term 'CLASH_RS_CI=true' is not recognized as a name of a cmdlet, function, script file, or executable
>      | program. Check the spelling of the name, or if a path was included, verify that the path is correct and try
>      | again.
> Error: Process completed with exit code 1.
> ```

i see...

Issue #321, opened 2024-03-15
shadow-tls implementation


VendettaReborn:
## Motivation 

The shadow-tls protocol may be one of safest protocol now? (I've used the shadow-tls-v3 for a long time, it's pretty stable)

## Current implementation in other repos

- original rust: https://github.com/ihciah/shadow-tls (monoio as the runtime)
- clash go impl: https://github.com/MetaCubeX/Clash.Meta/blob/Alpha/transport/shadowtls/
- another rust version: https://github.com/hsqStephenZhang/shadow-tls-tokio-client

## Status quo:

- in golang, the shadowtls client use the utls to hack the client hello packet
- in rust, since the TLS is rely on rustls, the monoio & tokio version of client both modify the rustls and change some code in the process of building client hello

## what we need to do

1. fork a rustls & tokio-rustls, do some changes similar to [shadow-tls-tokio-client](https://github.com/hsqStephenZhang/shadow-tls-tokio-client)
2. patch the Cargo.toml
3. add shadowtls's option in the shadowsocks's plugin
4. reuse the code of [shadow-tls-tokio-client](https://github.com/hsqStephenZhang/shadow-tls-tokio-client)

I had fully implement this protocol in my local environment and had it fully tested, however, i don't know what's the best way to **handle the dependency of the hacked** rustls & tokio-rustls: should i maintain the code by myself, or should we create a organization, and move the forked repos(rust-tun, netstack-lwip .etc ) to it?

ibigbug:
interesting. there's also discussions about this topic here https://github.com/refraction-networking/utls/issues/103

ibigbug:
> handle the dependency of the hacked
clash-rs is under an org account - https://github.com/Watfaq

if you don't mind, you can put the forked repo under this org.

hsqStephenZhang:
> interesting. there's also discussions about this topic here [refraction-networking/utls#103](https://github.com/refraction-networking/utls/issues/103)

Oh, I see. Actually, the patch that I've done is pretty similar to this one.

hsqStephenZhang:
> > handle the dependency of the hacked
> > clash-rs is under an org account - https://github.com/Watfaq
> 
> if you don't mind, you can put the forked repo under this org.

Got it.

VendettaReborn:
> > handle the dependency of the hacked
> > clash-rs is under an org account - https://github.com/Watfaq
> 
> if you don't mind, you can put the forked repo under this org.

Got it

[...]

Issue #190, opened 2023-11-24
transports


ibigbug:

betaxab:
Hi,

The current top priority is please add support for Hysteria2 and VLESS (Reality) protocols. These two protocols are more popular and accepted by people.

betaxab:
If possible, please implement SS2022(SIP022) as well.

The following link may be helpful:
https://github.com/shadowsocks/shadowsocks-org/issues/196
https://github.com/Shadowsocks-NET/shadowsocks-specs
https://github.com/shadowsocks/shadowsocks-rust
https://github.com/SagerNet/sing-shadowsocks
https://github.com/database64128/shadowsocks-go

itspig:
Please implement Snell protocol, Thanks!

leic4u:
vote to hysteria2 and vless+reality+vision

kumagaya68:
最新版本Nyanpasu内含0.2.0版本，也启动此内核，从配置文件可见订阅信息，但代理组没有节点信息。如何解决？

charley008:
please support vless+xhttp or reality?

X-49:
+1 for `xhttp`

PR #346, opened 2024-03-28
feat: support shadow-tls(v3 only)

VendettaReborn:
implement shadow-tls protocol's client(v3)

### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link

https://github.com/Watfaq/clash-rs/issues/321

### 💡 Background and solution

see the background mentioned in the issue. 

solution:
1. patch the rustls & tokio-rusls crates, support customed client id in the client hello
2. implement shadow-tls's client v3
3. add shadow-tls option in shadowsocks's plugins

Tests has been done in my ubuntu 22.04.

### 📝 Changelog

Support shadow-tls(v3 only)

### DIFF
--- a/clash_lib/src/proxy/converters/shadowsocks.rs
+++ b/clash_lib/src/proxy/converters/shadowsocks.rs
@@ -46,6 +46,15 @@ impl TryFrom<&OutboundShadowsocks> for AnyOutboundHandler {
                         .try_into()
                         .map(OBFSOption::V2Ray)
                         .ok(),
+                    "shadow-tls" => s
+                        .plugin_opts
+                        .clone()
+                        .ok_or(Error::InvalidConfig(
+                            "plugin_opts is required for plugin obfs".to_owned(),
+                        ))?
+                        .try_into()
+                        .map(OBFSOption::ShadowTls)
+                        .ok(),
                     _ => {
                         return Err(Error::InvalidConfig(format!(
                             "unsupported plugin: {}",
--- a/clash_lib/src/proxy/shadowsocks/shadow_tls/connector.rs
+++ b/clash_lib/src/proxy/shadowsocks/shadow_tls/connector.rs
@@ -1,0 +1,161 @@
+use std::io;
+use std::ptr::copy_nonoverlapping;
+use std::sync::Arc;
+
+use rand::Rng;
+
+use rand::distributions::Distribution;
+use tokio::io::{AsyncReadExt, AsyncWrite, AsyncWriteExt};
+use tokio_rustls::client::TlsStream;
+use tokio_rustls::rustls::ClientConfig;
+use tokio_rustls::TlsConnector;
+
+use crate::proxy::shadowsocks::ShadowTlsOption;
+use crate::proxy::AnyStream;
+
+use super::prelude::*;
+
+use super::stream::{ProxyTlsStream, VerifiedStream};
+use super::utils::Hmac;
+
+#[derive(Clone, Debug)]
+#[allow(unused)]
+pub struct Opts {
+    pub fastopen: bool,
+    pub sni: String,
+    pub strict: bool,
+}
+
+pub struct Connector {
+    pub password: String,
+    pub server_addr: String,
+    pub tls_config: ClientConfig,
+    pub connector: TlsConnector,
+}
+
+impl Connector {
+    fn connector() -> TlsConnector {
+        use crate::common::tls::GLOBAL_ROOT_STORE;
+
+        let tls_config = rustls::ClientConfig::builder()
+            .with_safe_defaults()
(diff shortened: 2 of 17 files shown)


VendettaReborn:
here is the shadow-tls's server startup script(via docker-compose), you may test or use it by running this script and copy&paste the output proxy config of this outbound.

https://gist.github.com/VendettaReborn/d8f8b546241e259f1cb18ce7b36b4106

clash_lib/src/proxy/shadowsocks/shadow_tls/connector.rs
```diff
@@ -0,0 +1,161 @@
...
+            .map(|s| (s.server_random, s.hmac.to_owned()));
+
+        if (!authorized || maybe_server_random_and_hamc.is_none()) && opts.strict {
+            tracing::warn!("V3 strict enabled: traffic hijacked or TLS1.3 is not supported, perform fake request");
+
+            tls.get_mut().0.fake_request = true;
+
+            let r = fake_request(tls).await;
+
+            return Err(io::Error::new(io::ErrorKind::Other, format!("V3 strict enabled: traffic hijacked or TLS1.3 is not supported, fake request, res:{:?}", r)));
```
ibigbug: do you not need to check the result before turning Error?

clash_lib/src/proxy/converters/shadowsocks.rs
```diff
@@ -46,6 +46,15 @@ impl TryFrom<&OutboundShadowsocks> for AnyOutboundHandler {
                         .try_into()
                         .map(OBFSOption::V2Ray)
                         .ok(),
+                    "shadow-tls" => s
```
ibigbug, line 49: mind adding a docker test for this?

VendettaReborn: Alright, but need some modification of docker test. Since the shadow-tls is a plugin of shadowsocks, we need to extend the docker test's ability by enabling it to start multiple images altogether

clash_lib/src/proxy/shadowsocks/shadow_tls/connector.rs
VendettaReborn: actually, even if the fake_request succeeds, this function shall also return error, so i just combined them together.
the original client code can be found in [shadow-tls native client](https://github.com/ihciah/shadow-tls/blob/dee3a5a819d6f56cfdd56c44a0d42be186c44238/src/client.rs#L272) 
but it's true that by returning the error with '?', we can make the code more readable.

clash_lib/src/proxy/converters/shadowsocks.rs
VendettaReborn: or we shall create a new docker images, including the binary of shadowsocks&shadow-tls. what do you think

clash_lib/src/proxy/shadowsocks/shadow_tls/connector.rs
ibigbug: i see. or just put a comment, the proto it self is a bit confusing here

clash_lib/src/proxy/converters/shadowsocks.rs
ibigbug: right. another option might be extending our tests to support building a local image and we make a Dockerfile for ss and the plugin.

this can be done in a separate PR later I guess.

VendettaReborn: i've extended the test utility and run multiple images at the same time as the preparation for the docker test. you can have a quick look, shall i merge it in this PR or in a separate one?

clash_lib/src/proxy/shadowsocks/shadow_tls/connector.rs
VendettaReborn: yep, i've fixed the logic and added some comments here

clash_lib/src/proxy/converters/shadowsocks.rs
ibigbug: yeah i think it's better to separate for easier review

VendettaReborn: alright

ibigbug (approved):
lgtm. plz resolve the conflicts

ibigbug (approved):

VendettaReborn:
mention roadmap https://github.com/Watfaq/clash-rs/issues/190

PR #370, opened 2024-04-12
Feat/udp docker test

VendettaReborn:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Performance optimization
- [x] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 💡 Background and solution

Currently, the docker test only covers TCP, but ignores UDP
So I added udp pingpong test and udp url_test(dns query) to cover it, and improved the coverage rate slightly

### 📝 Changelog

dev: support udp test in docker test
fix: fix datagram no `poll_flush` error in vmess & trojan & shadowsocks

### ☑️ Self-Check before Merge

⚠️ Please check all items below before requesting a reviewing. ⚠️

- [x] Doc is updated/provided or not needed
- [x] Changelog is provided or not needed

### coverage rate

the coverage rate is calculated by `cargo tarpaulin`.

- when we only have tcp test, the coverage rate under `proxy` directory is:  37%
- after, the coverage rate under `proxy` directory  is: 45%

PR #407, opened 2024-05-14
fix(testcase): fix wireguard testcase's timeout problem

VendettaReborn:
Test case: fix wireguard docker_test timeout error

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link

This [PR](https://github.com/Watfaq/clash-rs/pull/370) added udp test, and changed the latency test behavior, but after it, the wireguard's docker_test doesn't work anymore. 

### 💡 Background and solution

After some test, I found that it was caused by launching url test before the wireguard's docker image was full bootstrapped. 
So it can be simplify solved by adding a short sleep, but we can explore some more robust testing strategies, such as adding retry logic. But currently, it's not needed. 
Some other changes under `wg_config` was required for this fix, but they may cause some misunderstanding of `linuxserver/docker-wireguard`'s config, so i changed them for better readability.

### 📝 Changelog

fix(testcase): fix wireguard testcase


ibigbug (approved):
👍

PR #423, opened 2024-05-20
ci: enable SIMD instructions for x86_64 targets

Itsusinn:
### 🤔 This is a ...


Itsusinn:
https://en.wikipedia.org/wiki/SSE2
https://en.wikipedia.org/wiki/SSE3
https://en.wikipedia.org/wiki/Advanced_Vector_Extensions

Itsusinn:
https://zh.wikipedia.org/zh-cn/Sandy_Bridge%E5%BE%AE%E6%9E%B6%E6%A7%8B
https://zh.wikipedia.org/wiki/AES%E6%8C%87%E4%BB%A4%E9%9B%86

Itsusinn:
```
❯ elfx86exts clash-x86_64-unknown-linux-gnu
File format and CPU architecture: Elf, X86_64
MODE64 (call)
CMOV (cmova)
SSE1 (xorps)
SSE2 (movdqa)
BMI (tzcnt)
AES (aeskeygenassist)
AVX (vmovups)
NOVLX (vmovups)
AVX2 (vpbroadcastb)
SSSE3 (pshufb)
PCLMUL (pclmulqdq)
SSE41 (pblendw)
SSE3 (lddqu)
SHA (sha1rnds4)
BMI2 (shlx)
AVX512 (kmovw)
VLX (vpcmpltud)
ADX (adox)
3DNow (femms)
NOT64BITMODE (xchg)
Instruction set extensions used: 3DNow, ADX, AES, AVX, AVX2, AVX512, BMI, BMI2, CMOV, MODE64, NOT64BITMODE, NOVLX, PCLMUL, SHA, SSE1, SSE2, SSE3, SSE41, SSSE3, VLX
CPU Generation: Unknown
```
Seems SIMD is enabled by default on x86_64 targets.
But I wonder what will happen if old cpus use those binaries.

ibigbug (approved):

ibigbug:
Okay makes sense

VendettaReborn:
It sounds like a easy-to-obtain improvement, but I'd like to share [Rob pike's programming principles](https://[users.ece.utexas.edu/~adnan/pike.html](https://users.ece.utexas.edu/~adnan/pike.html))
from my view, the real problem is, we even don't know where the bottleneck is, and we don't have any measures except the performance boost in theory.
Btw, I'm also thinking about ways to compare clash-rs's performance with clash.meta and sing-box. It would be fun. Maybe we can move to discussions.

Itsusinn:
> from my view, the real problem is, we even don't know where the bottleneck is, and we don't have any measures except the performance boost in theory.

We do need some performance characterization methods.But I think the performance of clash-rs is limited to user actual network condition.(CN2, BGP etc.) The perf test on local loopback could never be reach in realworld.

VendettaReborn:
that's true, the network condition between test machines must be stable enough, in another words, they'd better exist under the same LAN

Itsusinn:
Close in favor of https://github.com/RustCrypto/AEADs/issues/427#issuecomment-1142704771

Issue #421, opened 2024-05-20
Jaeger: no http client


VendettaReborn:
when enabling jaeger via "JAGER_ENDPOINT", i encountered the error:
`failed to setup logging: Exporter otlp encountered the following error(s): no http client, you must select one from features or provide your own implementation`

it seems that `otlp_exporter` missed the calling of `with_http_client`, there are two ways of fixing this:
1. use grpc client instead of http client (tonic has already been a dependency in other places, so won't lead to extra imports)
2. add deps of `opentelemetry_http`, create a client and pass it to  `with_http_client`

the former one is simplier, but the jaeger's startup command has to be changed into: 
`docker run -d -p16686:16686 -p4317:4317 -p4318:4318 -e COLLECTOR_OTLP_ENABLED=true jaegertracing/all-in-one:latest` to use the grpc port 4317

ibigbug:
Yeah it's a known issue. https://github.com/open-telemetry/opentelemetry-rust/issues/1659#event-12795331511

Maybe time to move to the grpc transport.

PR #427, opened 2024-05-22
fix: fix NoHttpClient error in opentelemetry-otlp, use grpc instead

VendettaReborn:
### 🤔 This is a ...

- [x] Bug fix

### 🔗 Related issue link

https://github.com/Watfaq/clash-rs/issues/421

### 💡 Background and solution

use grpc instead of http client in opentelemetry-otlp

### 📝 Changelog

switch to grpc client for jaeger


ibigbug (approved):
tks!

Issue #426, opened 2024-05-22
buf.len() must fit in remaining()


MrZLeo:
```
2024-05-22T02:27:01.861925Z ERROR dispatch_stream:copy_bidirectional: clash_lib/src/lib.rs:158: panic hook: PanicInfo { payload: Any { .. }, message: Some(buf.len() must fit in remaining()), location: Location { file: "clash_lib/src/proxy/shadowsocks/simple_obfs/http.rs", line: 109, col: 33 }, can_unwind: true, force_no_backtrace: false } outbound_name="🔎 Google"
```

VendettaReborn:
yeah, there is a bug in shadowsocks's obfs impl, the check before `put_slice` is missed, we will fix it.

VendettaReborn:
try [2c5f937](https://github.com/Watfaq/clash-rs/commit/2c5f9377a88bfeaf2bc3fc9e196c04ce8273d056), the binaries can be found [here](https://github.com/Watfaq/clash-rs/actions/runs/9211067069)

MrZLeo:
I used master branch which looks like conver 2c5f937 commit, but another error happened:

```
2024-05-27T11:05:44.383987Z ERROR dispatch_stream:copy_bidirectional: clash_lib/src/lib.rs:158: panic hook: PanicInfo { payload: Any { .. }, message: Some(source slice length (5646) does not match destination slice length (0)), location: Location { file: "clash_lib/src/proxy/shadowsocks/simple_obfs/http.rs", line: 130, col: 46 }, can_unwind: true, force_no_backtrace: false } outbound_name="🔎 Google"
```

VendettaReborn:
got it, i will take a look

VendettaReborn:
@MrZLeo i've added docker tests for obfs in [this PR](https://github.com/Watfaq/clash-rs/pull/437). and we are certain that obfs's http & tls modes are usable.

PR #435, opened 2024-05-28
fix: use put_slice instead of clone_from_slice in httpobfs

VendettaReborn:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link

https://github.com/Watfaq/clash-rs/issues/426

### 💡 Background and solution

BG: the issue has been fixed once, after that, i did do some in-the-real-world tests, but some situation is not triggered/covered. and it seems that the bugfix has some problems: 

1. clone_from_slice is misused, and will cause panic.
2. the buffered fist response shall be checked at the beginning of poll_read

so i mainly refered to [clash.meta's impl](https://github.com/MetaCubeX/mihomo/blob/48e481d0a22302c215878bd53e0b81547f4f45de/transport/simple-obfs/http.go#L28) this time. 

### additional thoughts

I still couldn't trigger the modified code in my computer, but the code should be alright this time. 
But i got to think, in each protocol's impl, there may exists some situations that is not covered by our docker test. Shall we figure out some approach to verify each protocol's internal states, since the `ByteBuf` and `ReadBuf`'s usages are fallible

### 📝 Changelog

fix: fix panic in http simple-obfs in ss(again).

### DIFF
--- a/clash_lib/src/proxy/shadowsocks/simple_obfs/http.rs
+++ b/clash_lib/src/proxy/shadowsocks/simple_obfs/http.rs
@@ -90,21 +90,20 @@ impl AsyncRead for HTTPObfs {
         buf: &mut tokio::io::ReadBuf<'_>,
     ) -> std::task::Poll<std::io::Result<()>> {
         let pin = self.get_mut();
-
-        if pin.first_response {
-            // as long as the buffer is not empty, we should return the data in the buffer first
-            if !pin.read_buf.is_empty() {
-                let to_read = std::cmp::min(buf.remaining(), pin.read_buf.len());
-                if to_read == 0 {
-                    assert!(buf.remaining() > 0);
-                    return std::task::Poll::Pending;
-                }
-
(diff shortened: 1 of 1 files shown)


ibigbug:
yeah i recall this happened before in somewhere that has been fixed. it would be great if the slice boundary check can be done at compile time instead of runtime panic.

maybe we can do some fuzzing around it.

found an old topic similar to this scenario https://www.reddit.com/r/rust/comments/ctlmdk/comment/exnq0fi/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

maybe something we can borrow from.

ibigbug (approved):

PR #437, opened 2024-05-29
fix: fix state machine logic error in obfs's tls impl; add docker tests for obfs

VendettaReborn:
### 🤔 This is a ...

- [x] Bug fix
- [x] Test Case

### 🔗 Related issue link

https://github.com/Watfaq/clash-rs/issues/426

### 💡 Background and solution

add docker test support for obfs(only on amd64), and found the bug in obfs's tls impl, and have it fixed

### 📝 Changelog

fix: fix bug in shadowsocks obfs's tls mode


ibigbug (approved):
nice!

PR #228, opened 2023-12-26
smoltcp

ibigbug:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Performance optimization
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link

#208 

blocked by https://github.com/smoltcp-rs/smoltcp/issues/856

### ☑️ Self-Check before Merge

⚠️ Please check all items below before requesting a reviewing. ⚠️

- [ ] Doc is updated/provided or not needed
- [ ] Changelog is provided or not needed

Issue #208, opened 2023-12-07
Use somltcp


ibigbug:

cavivie:
@ibigbug I think I might be of some help here, as I recently stabilized the netstack-smoltcp crate.

cavivie:
Well, I misunderstood you. It seems that you mean tcp/udp.

ibigbug:
yeah that's waht i meant

ibigbug:
@cavivie were you planning to  creat a PR for this?

cavivie:
I'd love to, but before that, can you add more about tun in ss, including unresolved problems and known problems and goals.

ibigbug:
I think the main goal is to replace the lwip w smoltcp, for performance boost, see https://github.com/Watfaq/netstack-smoltcp/blob/main/benchmark.md

And adopting pure rust impls over foreign C bindings. 

This is only for the tun input, not much todo with ss. 

We already use smoltcp in the user space wirefuard, it consolidates the deps which is another good reason too. 

And thanks for you work.

cavivie:
Thank you for your description. I think I will take time to integrate it recently.

VendettaReborn:
@cavivie @ibigbug i added some soem auto detect interface feature for rust-tun, and here is the example code: https://github.com/VendettaReborn/netstack-smoltcp/blob/f89b9e9b2e56a6abf1b2be9290e4cf63c7be3e11/examples/forward.rs#L142 . Basically, it's just a code version of route commands. And i've tested it on linux,macos & windows.

Shall we extract the logic of auto-detect interface, auto setup routes into a independent crate?  just like sing-tun?

Itsusinn:
> Shall we extract the logic of auto-detect interface, auto setup routes into a independent crate?  just like sing-tun? 
i would like call it `watfaq-tun` : )

VendettaReborn:
any updates?

Itsusinn:
> any updates?

最近挺忙的，大概到六月初才有时间

cavivie:
Ah, ah, first of all, I'm sorry for the reason why I've been busy recently. I want to disassemble the current steps here:
1. 
> Shall we extract the logic of auto-detect interface, auto setup routes into an independent crate? just like sing-tun?

Are there any disassembled independent clates here? I want to do some work based on this.

2.
BTW, netstack-smoltcp may have some things that need to be changed. Can I think of anything else you need to add besides some refactoring of the udp part?

[...]

PR #456, opened 2024-06-12
refactor(tun): use libtun

Itsusinn:
### 🤔 This is a ...

- [x] Refactoring

### 🔗 Related links
#208 

#362
#228
### Contents
- [x] Replace `netstack-lwip` with `netstack-smoltcp`
- [ ] Auto route

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -578,29 +578,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bindgen"
-version = "0.69.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a00dc851838a2120612785d195287475a3ac45514741da670b735818822129a0"
-dependencies = [
- "bitflags 2.5.0",
- "cexpr",
- "clang-sys",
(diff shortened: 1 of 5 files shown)


ibigbug:
@VendettaReborn @cavivie thoughts?

seems this PR uses tun2 https://github.com/Itsusinn/libtun/blob/e8db4880bd5cca35930602222e7cfa549533f1f1/Cargo.toml#L8C8-L8C25

cavivie:
I think we should migrate to tun2. Is the reason why we continue to use tun too much incompatible? We can abstract tun and netstack-lwip to libtun, and clash_lib does not directly depend on tun crate.

VendettaReborn:
> I think we should migrate to tun2. Is the reason why we continue to use tun too much incompatible? We can abstract tun and netstack-lwip to libtun, and clash_lib does not directly depend on tun crate.

agreed

ibigbug:
Yeah sounds good to me

xmh0511:
We can use `tun2`, `ipstack`, and `net-route` to achieve the target.

PR #458, opened 2024-06-16
feat: try to resolve default gateway interface on tun inbound traffic to avoid route loop

ibigbug:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [x] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)


ibigbug:
now i know why you guys keep talking about auto-tun - i never had the issue before on macOS with a single IP route tun mode, i.e. something like:

```
sudo ip route add 10.0.0.11/32 dev utun1989
```

but today i realized it causes traffic loop on linux for some reason( i forgot if i ever faced this issue before, maybe you guys have more experience with this matter)

i'm just adding a simple dirty hack for tun to work with an attempt to resolve default iface for tun traffic, feel free to add your opinions

VendettaReborn:
yeh, to detect the default interface is necessary, and we should also detect if there exists some loop and downgrade the service.

ibigbug:
Agreed. I was a bit surprised this was never an issue when I was developing mostly on non Linux environment and adding a single route just worked fine.

VendettaReborn:
i haven't faced this issue too, maybe the network condition is stable? if the default interface changes when network switches, something bad could happen(according to clash-meta's issues)

ibigbug:
i should only have one single interface plus a lo on that VM, before the utun was added. and it somehow the request to the newly added target went back to the utun again(and again) and didn't go through the default interface. which kind of makes sense as we told the routing table to route the target through utun.

the current hack should workaround the issue to force select a outbound interface for sessions from utun. we should probably looking into a more sophisticated solution as you mentioned before

VendettaReborn:
have you test the code when the interface is down? since the default interface is not usable, the outbound traffic probably will go through the utun again and again, and that could leak to OOM.

PR #465, opened 2024-06-19
refactor: tuic

Itsusinn:
cherry pick of #363

### DIFF
--- a/clash_lib/src/proxy/converters/tuic.rs
+++ b/clash_lib/src/proxy/converters/tuic.rs
@@ -43,8 +43,7 @@ impl TryFrom<&OutboundTuic> for AnyOutboundHandler {
             heartbeat_interval: Duration::from_millis(s.heartbeat_interval.unwrap_or(3000)),
             reduce_rtt: s.reduce_rtt.unwrap_or(false) || s.fast_open.unwrap_or(false),
             request_timeout: Duration::from_millis(s.request_timeout.unwrap_or(4000)),
-            idle_timeout: VarInt::from_u64(s.request_timeout.unwrap_or(4000))
-                .unwrap_or(VarInt::MAX),
+            idle_timeout: Duration::from_millis(s.request_timeout.unwrap_or(4000)),
             congestion_controller: s
                 .congestion_controller
                 .clone()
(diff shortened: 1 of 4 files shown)


clash_lib/src/proxy/converters/tuic.rs
```diff
@@ -37,7 +42,9 @@ impl TryFrom<&OutboundTuic> for AnyOutboundHandler {
                 .unwrap_or_default(),
             heartbeat_interval: Duration::from_millis(s.heartbeat_interval.unwrap_or(3000)),
             reduce_rtt: s.reduce_rtt.unwrap_or(false) || s.fast_open.unwrap_or(false),
-            request_timeout: Duration::from_millis(s.request_timeout.unwrap_or(8000)),
+            request_timeout: Duration::from_millis(s.request_timeout.unwrap_or(4000)),
+            idle_timeout: VarInt::from_u64(s.request_timeout.unwrap_or(4000))
+                .unwrap_or(VarInt::MAX),
```
ibigbug: could you use Duration here too?

Also if it failed to pass we should use the same default value 4000 rather than MAX?

clash_lib/src/proxy/tuic/handle_task.rs
```diff
@@ -141,7 +144,7 @@ impl TuicConnection {
         }
 
         match self.inner.heartbeat().await {
-            Ok(()) => tracing::trace!("[heartbeat]"),
+            Ok(()) => tracing::debug!("[heartbeat]"),
```
ibigbug: maybe add more context to this line, such as [quic heartbeat] with other possible information (remote endpoint etc). it's hard to track where this is originated when logs are verbose

clash_lib/src/proxy/tuic/mod.rs
```diff
@@ -144,34 +143,46 @@ impl Handler {
...
+            CongestionControl::Bbr => {
+                transport_config.congestion_controller_factory(Arc::new(BbrConfig::default()))
+            }
+        };
+
+        quinn_config.transport_config(Arc::new(transport_config));
         // Try to create an IPv4 socket as the placeholder first, if it fails, try IPv6.
+        // If it dont match server's ipv4/ipv6, rebind is needed
         let socket =
             UdpSocket::bind(SocketAddr::from((Ipv4Addr::UNSPECIFIED, 0))).or_else(|err| {
```
ibigbug, line 169: i have a sense that the iOS issue in the group is somehow related to this bind iirc iOS doesn't allow bind to 0.0.0.0. I will take a look again and might change here

ibigbug (changes requested):
lgtm with minor comments

ibigbug (approved):

PR #555, opened 2024-08-24
feate: linux-musl target,docker image,add rc4-md5 cipher

litcc:
### 🤔 This is a ...

- [x] New feature
- [x] Other (CI script)

### 🔗 Related issue link

None

### 💡 Background and solution

1. I need to deploy the clash proxy in a containerized environment.
2. Some proxy service providers only support rc4-md5 cipher.

### 📝 Changelog

Tweaked CI build to add x86_64-unknown-linux-musl target
Added docker image based on x86_64-unknown-linux-musl package.
Added rc4-md5 cipher parsing.

**You need to add two action secrets: DOCKER_USERNAME and DOCKER_PASSWORD.**

### DIFF
--- a/.github/Dockerfile
+++ b/.github/Dockerfile
@@ -1,5 +1,7 @@
 FROM alpine:latest
-COPY ./clash-rs/clash-x86_64-unknown-linux-musl /usr/bin/clash
+# Define an ARG for the target architecture
+ARG TARGETARCH
+COPY ./clash-rs/clash-${TARGETARCH} /usr/bin/clash
 # The yq library installed here is used to rewrite the config.yaml configuration file for clash, merge it, and other related operations.
 RUN apk update && apk add --no-cache -f yq && mkdir -p /root/.config/clash/
 WORKDIR /root
--- a/.github/Dockerfile
+++ b/.github/Dockerfile
@@ -1,6 +1,7 @@
-# A base image built by an ops bigwig, open source
-FROM wener/base:latest
-COPY ./clash-rs/clash-x86_64-unknown-linux-musl /usr/local/bin/clash
-RUN apk add --no-cache -f yq && mkdir -p /root/.config/clash/
+FROM alpine:latest
+COPY ./clash-rs/clash-x86_64-unknown-linux-musl /usr/bin/clash
+# The yq library installed here is used to rewrite the config.yaml configuration file for clash, merge it, and other related operations.
+RUN apk update && apk add --no-cache -f yq && mkdir -p /root/.config/clash/
 WORKDIR /root
-CMD ["/usr/local/bin/clash","-d", "/root/.config/clash/"]
+ENTRYPOINT [ "/usr/bin/clash" ]
+CMD [ "-d", "/root/.config/clash/" ]
--- a/.github/Dockerfile
+++ b/.github/Dockerfile
@@ -1,6 +1,6 @@
 # A base image built by an ops bigwig, open source
 FROM wener/base:latest
-COPY ./clash-rs-upx /usr/local/bin/clash
+COPY ./clash-rs/clash-x86_64-unknown-linux-musl /usr/local/bin/clash
 RUN apk add --no-cache -f yq && mkdir -p /root/.config/clash/
 WORKDIR /root
 CMD ["/usr/local/bin/clash","-d", "/root/.config/clash/"]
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -272,3 +280,58 @@ jobs:
...
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Download binaries
+        uses: actions/download-artifact@v4
+        with:
+          name: x86_64-unknown-linux-musl
+          path: ./clash-rs
+      - name: Install UPX
+        uses: crazy-max/ghaction-upx@v3
+        with:
+          install-only: true
```
Itsusinn: Please remove any UPX parts, because it might increase memory usage.

.github/workflows/ci.yml
```diff
@@ -272,3 +280,58 @@ jobs:
           files: |
             packages/*
             LICENSE
+  docker-image:
```
Itsusinn, line 289: Please change entire job according to https://github.com/Itsusinn/tuic/blob/dev/.github%2Fworkflows%2Fci.yml#L250-L305

.github/Dockerfile
```diff
@@ -0,0 +1,6 @@
+# A base image built by an ops bigwig, open source
```
Itsusinn: check https://github.com/Itsusinn/tuic/blob/dev/.github%2FDockerfile

.github/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM alpine:latest
+COPY ./clash-rs/clash-x86_64-unknown-linux-musl /usr/bin/clash
```
Itsusinn: Will this fail when building arm image？

litcc: The alpine base image exists as an arm build, which is reasonably fine.
It is tested to be able to package the image properly, but the actual availability needs to be tested on an arm machine, and I don't have an arm machine at the moment.

Itsusinn: Alpine is fine. I mean you copied x86_64 clash-rs into arm image.

litcc: You're right. I'll check.

.github/workflows/ci.yml
```diff
@@ -272,3 +278,67 @@ jobs:
...
```
ibigbug, line 289: can we only do this on `release`?

ibigbug (changes requested):
lgtm only small note

.github/workflows/ci.yml
litcc: As I understand it, master should be a dev-like branch in that project, and commits that are tagged exist to be releases, right? Am I understanding you correctly? Because I don't see a dev branch associated with it.

litcc: > can we only do this on `release`?

A nightly release doesn't require a build image either?

Itsusinn: We can, but why?

Itsusinn (dismissed):

.github/workflows/ci.yml
```diff
@@ -272,3 +286,71 @@ jobs:
...
+      - name: Build and push release
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          file: .github/Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: ${{ env.TAG_VERSION }},${{ env.TAG_LATEST }}
+
+      - name: Delete all images from package without tags
```
Itsusinn: Remove this if we only build images on `release`

.github/workflows/ci.yml
ibigbug: ok i might be a bit unclear - i saw this PR had a docker image built(that failed with no docker creds, which I'll add), which I don't think we need, before merging?

ibigbug:
image

litcc:
> image


The latest docker builds use `secrets.GITHUB_TOKEN` which is automatically injected by action and does not need to be generated separately.

ibigbug:
For docker hub or GitHub docker registry?

litcc:
> For docker hub or GitHub docker registry?

Currently only for GitHub image registry (ghcr.io), do I need to support docker.io?

ibigbug:
nah ghcr is good enough for now

ibigbug (approved):
tysm!

Issue #559, opened 2024-08-27
Implement auto route for tun inbound


ibigbug:
## Tenets

* clash-rs should handle traffic on given routes when tun inbound is enabled
* users should be able to tell the routes they want to proxy over tun
* should work on Windows/Linux/macOS
* we should try our best to resolve routes failures and give as much details when things don't work

## Work Items
- [ ] #228 
- [ ] #456 
- [x] #541 
- [x] #552 
- [x] #396
- [ ] #579 
- [x] platform
    - [x] Windows
      * doesn't work - https://learn.microsoft.com/en-us/windows/win32/api/rtmv2/nf-rtmv2-rtmaddnexthop
      * use built in command instead
    - [x] Linux - https://www.wireguard.com/netns/ either fwmark or ns
    - [x] macOS - assumes `route` command is builtin
- [ ] ipv6 - blocked by #579 

## Example config

```yaml
tun:
  enable: true
  device-id: "dev://utun1989"
  routes:
    - 0.0.0.0/1
    - 128.0.0.0/1
```

## Implementation

- [x] Windows - https://github.com/Watfaq/clash-rs/pull/561
- [x] macOS - #592 
- [x] Linux - #594

PR #561, opened 2024-08-29
feat(tun): windows auto-route

ibigbug:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [x] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link

#559 

### 💡 Background and solution

https://learn.microsoft.com/en-us/windows/win32/rras/add-and-update-routes-using-rtmaddroutetodest

### 📝 Changelog

```
tun:
  enable: true
  device-id: "dev://utun1989"
  route-all: false
  routes:
    - 0.0.0.0/1
    - 128.0.0.0/1
```

### DIFF
--- a/clash_lib/src/proxy/tun/inbound.rs
+++ b/clash_lib/src/proxy/tun/inbound.rs
@@ -1,7 +1,13 @@
 use super::{datagram::TunDatagram, netstack};
-use std::{net::SocketAddr, sync::Arc};
+use std::{
+    net::{Ipv4Addr, SocketAddr},
+    sync::Arc,
+    u32,
+};
 
 use futures::{SinkExt, StreamExt};
+use ipnet::IpNet;
+use network_interface::NetworkInterfaceConfig;
 use tracing::{debug, error, info, trace, warn};
 use tun::{Device, TunPacket};
 use url::Url;
@@ -10,7 +16,11 @@ use crate::{
     app::{dispatcher::Dispatcher, dns::ThreadSafeDNSResolver},
     common::errors::{map_io_error, new_io_error},
     config::internal::config::TunConfig,
-    proxy::{datagram::UdpPacket, utils::get_outbound_interface},
+    proxy::{
+        datagram::UdpPacket,
+        tun::routes::add_route,
+        utils::{get_outbound_interface, OutboundInterface},
+    },
     session::{Network, Session, SocksAddr, Type},
     Error, Runner,
 };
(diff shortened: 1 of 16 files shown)


clash_lib/src/proxy/tun/inbound.rs
```diff
@@ -177,14 +186,56 @@ pub fn get_runner(
...
+                "route_all is enabled, all traffic will be routed through the tun \
+                 interface"
+            );
+            let default_routes = vec![
+                IpNet::new(std::net::IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), 1)
+                    .unwrap(),
+                IpNet::new(std::net::IpAddr::V4(Ipv4Addr::new(128, 0, 0, 0)), 1)
+                    .unwrap(),
+            ];
+            for r in default_routes {
```
VendettaReborn: shall we make this a two step operation of build_route**s** & add_route**s**, thus is can be separated into the `routes` module

clash_lib/src/proxy/tun/routes/windows.rs
```diff
@@ -0,0 +1,190 @@
...
+        error!("failed to add route: {}", err);
+        Err(new_io_error(err.to_string().as_str()))
+    }
+}
+
+/// Add a route to the routing table.
+/// https://learn.microsoft.com/en-us/windows/win32/rras/add-and-update-routes-using-rtmaddroutetodest
+/// FIXME: figure out why this doesn't work https://stackoverflow.com/questions/43632619/how-to-properly-use-rtmv2-and-rtmaddroutetodest
+#[allow(dead_code)]
+pub fn add_route_that_does_not_work(
```
VendettaReborn, line 55: this is how `net-route` handled the routing table: https://github.com/johnyburd/net-route/blob/f83a447bd13404225c8a78db71af208ee8e7885c/src/platform_impl/windows.rs#L178. it seems that they took a different approach

Itsusinn:
实测能通吗？windows api那部分实在是看不懂

Itsusinn:
别修了吧，感觉没啥用，微软的文档都不全

Itsusinn (approved):

Itsusinn (approved):

VendettaReborn (approved):

clash_lib/src/proxy/tun/routes/windows.rs
ibigbug: yeah i checked their code they used a different set of API. which makes sense. maybe I'll switch to using the crate for this too when I do the macos one

clash_lib/src/proxy/tun/inbound.rs
ibigbug: i'm hoping the 0/1 and 128/1 can work on all platforms so the build_route will just be the same 2 entries

Issue #597, opened 2024-09-18
[feature] add a `checksum` file to help updater validate the download content


greenhat616:
FYI Github have removed the `md5` signature header. Ref: https://github.com/orgs/community/discussions/23512#discussioncomment-10013377

So, we should upload a checksum file, `.sha256` etc, to make the integrity of file checkable.

For example, we can do it in ci:
```yaml
- name: Calc the archive signature
        shell: pwsh
        run: Get-FileHash nyanpasu-service-${{ matrix.profile.target }}.zip -Algorithm SHA256 | Format-List > nyanpasu-service-${{ matrix.profile.target }}.zip.sha256
```

```yaml
 - name: Calc the archive signature
        run: sha256sum nyanpasu-service-${{ matrix.profile.target }}.tar.gz > nyanpasu-service-${{ matrix.profile.target }}.tar.gz.sha256
```

Ref: https://github.com/libnyanpasu/nyanpasu-service/blob/main/.github/workflows/build.yml

Maybe I can draft a pr to solve it later.

Itsusinn:
There would be a `sha256sum.txt` , possibly.
It looks like
```
162b0b32f02482d5aca0a7c93dd03ceac3acd7e410a5f18f3fb990fc958ae0df6f32233b91831eaf99ca581a8c4ddf9c8ba315ac482db6d4ea01cc7884a635be  file1
46f93691eef6d2e005b4d847f779740f181ca7facf3136f0bc095e543372b14e997d47e52257d90284758ac80fa2ba4ef68bab751deb03f96746f877847b3251  file2
```

Issue #596, opened 2024-09-18
[feature] nightly alpha build add version info


greenhat616:
* use alpha version, such as `0.3.2-alpha.hash` to build
* add a `version.txt` file to let updater could check the updates
  * if use `update-tag` action, the hash of release in the page is updated, but the result of Github api is not.
  * the api of github have a request limit, and the file is not.

Maybe i can draft a pr to solve it later

Itsusinn:
For nightly release，I would prefer add a `version.txt` alongs with binaries.
The content of `version.txt` might be:
- Hash of commit
- Time-stamp.

PR #599, opened 2024-09-19
ci: add sha256sum and version.txt

Itsusinn:


Itsusinn:
#596 #597

Itsusinn:
@greenhat616 If you have any alternative requirement，please tell me

greenhat616:
Maybe we could use the `cargo set-version` from [cargo-edit](https://crates.io/crates/cargo-edit) to make the consistent with the `version.txt`?
And here is the spec of semver: <https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions>
So that `x.y.z-tag+buildinfo` is vaild for semver.

If use are inconsistant with the version.txt. The updater should create a independent lockfile to fix version by file chksum, which is not implemented yet now.

Here is the mihomo alpha channel output:
```
Mihomo Meta alpha-3676d1b windows amd64 with go1.23.1 Wed Sep 18 14:44:10 UTC 2024
Use tags: with_gvisor
```

It is consistent with the content of `version.txt`.
For manifest v1, we parsed the version, using it compare with the remote `version.txt`, and do the update logic.

Itsusinn:
> we might use the `cargo set-version` from [cargo-edit](https://crates.io/crates/cargo-edit) to make the consistent with the `version.txt`?

Doing so will crate tons of commits by bot（kinda of spam i have to say）.

So no.

> If the program version are inconsistant with the `version.txt`. The updater are supposed to create a independent lockfile to fix version by file chksum, which is not implemented yet now.

Is the `program version` you pointed out here meaning cli's `clash-rs -v` output？

If so，that's doable.

greenhat616:
> we might use the `cargo set-version` from [cargo-edit](https://crates.io/crates/cargo-edit) to make the consistent with the `version.txt`?

I think we do not need to commit the changes after modify the version for alpha build, because we don't release a real alpha tag in github like other well-known repos do.

> > If the program version are inconsistant with the `version.txt`. The updater are supposed to create a independent lockfile to fix version by file chksum, which is not implemented yet now.
> 
> Is the `program version` you pointed out here meaning cli's `clash-rs -v` output？
> 
> If so，that's possible to do.

Yes

Itsusinn:
> I think we do not need to commit the changes after modify the version for alpha build, because we don't release a real alpha tag in github like other well-known repos do.

Very good thinking!

Now new version (.txt) format in my mind: `0.x.y-alpha+sha.000000`

greenhat616:
> > I think we do not need to commit the changes after modify the version for alpha build, because we don't release a real alpha tag in github like other well-known repos do.
> 
> Very good thinking!
> 
> Now new version (.txt) format in my mind: `0.x.y-alpha+sha.0000000`

Yeah. That's enough.

ibigbug:
Yeah let's also change the cli version output for the nightly release to make it aligned

greenhat616:
> > I think we do not need to commit the changes after modify the version for alpha build, because we don't release a real alpha tag in github like other well-known repos do.
> 
> Very good thinking! But it still makes action more complex, we have to do these things **in every target**:
> 
>     * Parse current version from `Cargo.toml`
> 
>     * Get short SHA from `git`
> 
>     * Install `cargo-edit` by `cargo install` (another few minutes)
> 
>     * Join vars, and `cargo set new-version`
> 
> 
> I would prefer add a environment called `VERSION_OVERRIDE` And in cli process:
> 
> ```rust
> let version = match env!(VERSION_OVERRIDE) {
>     Some(v) -> v
>     None -> env!(CARGO_PKG_VERSION)
> }
> ```
> 
> Now new version (.txt) format in my mind: `0.x.y-alpha+sha.0000000`

Yeah. If we use the `cargo-edit`, we'd better use [cargo-binstall](https://github.com/cargo-bins/cargo-binstall?tab=readme-ov-file) to install it.

And the build envs var is a good way too.

clash/src/main.rs
```diff
@@ -31,13 +31,27 @@ struct Cli {
...
+    version: bool,
     #[clap(short, long, help = "Additinally log to file")]
     log_file: Option<String>,
 }
 
 fn main() {
     let cli = Cli::parse();
+
+    if cli.version {
+        println!("{}", env!("CLASH_VERSION_OVERRIDE"));
```
greenhat616: could follow the current style like below?

```shell
clash-rs 0.3.2
```

greenhat616 (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -377,13 +382,20 @@ jobs:
           mv ./clash-rs/clash-x86_64-unknown-linux-musl ./clash-rs/clash-amd64
           mv ./clash-rs/clash-aarch64-unknown-linux-musl ./clash-rs/clash-arm64
 
+      - name: Log in to Container Registry
```
ibigbug, line 385: is this intentional?

clash/build.rs
```diff
@@ -0,0 +1,14 @@
+#![feature(let_chains)]
+fn main() {
+    println!("cargo:rerun-if-env-changed=CARGO_PKG_VERSION");
```
ibigbug, line 3: this wouldn't rebuild on master when only ref changes but not pkg version right?

.github/workflows/ci.yml
Itsusinn: Yes，I about to refactor docker workflow，but failed. Probably another pr

clash/build.rs
Itsusinn: right, But for now it only cares package version. If someday we have extra requirements, it's not late to change it then.

ibigbug: But you'd want to rebuild for every master commit when version doesn't change?

Itsusinn: Will rebuild happen when actions are traiggered? 

I am a little not sure for now, what should I change this into?

ibigbug: Actually it should be fine. Any code change on master should trigger a rebuild anyways.

ibigbug (approved):

Itsusinn:
It seems broken......

PR #615, opened 2024-09-30
feat(tproxy): support tproxy on Linux

ibigbug:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link

#73 

### 💡 Background and solution

largely based on and supercedes #343 

without iptables integration

https://github.com/kristrev/tproxy-example/blob/master/README.md
https://powerdns.org/tproxydoc/tproxy.md.html

### DIFF
--- a/clash_lib/src/proxy/tproxy/mod.rs
+++ b/clash_lib/src/proxy/tproxy/mod.rs
@@ -1,11 +1,13 @@
-use crate::proxy::InboundListener;
+use std::{net::SocketAddr, sync::Arc};
+
+use crate::{app::dispatcher::Dispatcher, proxy::InboundListener};
 use async_trait::async_trait;
-use std::net::SocketAddr;
 
 use tracing::warn;
 
 pub struct Listener {
     addr: SocketAddr,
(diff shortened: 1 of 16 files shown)


clash_lib/src/proxy/tproxy/mod.rs
```diff
@@ -15,26 +17,202 @@ impl Drop for Listener {
...
+    let closer = dispatcher.dispatch_datagram(sess, Box::new(udp_stream));
+
+    // dispatcher -> tproxy
+    let responder = socket.clone();
+    let fut1 = tokio::spawn(async move {
+        while let Some(pkt) = l_rx.recv().await {
+            trace!("tproxy <- dispatcher: {:?}", pkt);
+
+            // remote -> local
+            match responder
```
ibigbug, line 134: @VendettaReborn we don't need to create a new response_transparent socket here and can just use the original listen socket to response

Issue #563, opened 2024-08-31
[Features] Load-balance improvements


ObcbO:
1. Add new strategy "sticky-sessions" https://github.com/MetaCubeX/mihomo/blob/Alpha/adapter/outboundgroup/loadbalance.go#L189
2. Control the node selection range in the configuration to allow only the top few nodes with the lowest latency or those with latency below a specified threshold, to avoid selecting nodes with high latency.

hsqStephenZhang:
@Noctiro try https://github.com/Watfaq/clash-rs/commit/0587fa8a54

PR #673, opened 2025-01-15
enhance(load balance): add StickySession load balance strategy

VendettaReborn:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [x] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link

enhancement in (https://github.com/Watfaq/clash-rs/issues/563)

### 💡 Background and solution

refer to [clash.meta's](https://github.com/MetaCubeX/mihomo/blob/1c5f4a3ab12ae748cf85b1f14e1841f3b570e9c3/adapter/outboundgroup/loadbalance.go#L189) solution

strategy details: 
- use the combination of src and dst ip addresses as the hash key
- cache the chosen proxy's index in a lru_cache with 10m expire duration
- always check the aliveness(by default is alive) before select the proxy

tests: 
- create `NoopResolver` and `NoopOutboundHandler` to create ProxyManager freely (shall i separate it into another PR?)

### 📝 Changelog

enhance(load balance): support sticky session strategy.

unitests in `loadbalance/helpers.rs`


clash_lib/src/proxy/group/loadbalance/helpers.rs
```diff
@@ -63,3 +80,195 @@ pub fn strategy_consistent_hashring() -> StrategyFn {
         )))
     })
 }
+
+#[cfg(test)]
```
ibigbug, line 84: can we move this into mod tests

VendettaReborn: ok

clash_lib/src/proxy/group/loadbalance/helpers.rs
```diff
@@ -63,3 +80,195 @@ pub fn strategy_consistent_hashring() -> StrategyFn {
...
+                        }
+                        return Ok(proxy.clone());
+                    }
+                }
+                // the cached proxy is dead, change the key by a new timestamp and
+                // try again
+                index = jump_hash(key + timestamp(), buckets) as usize;
+            }
+            // TODO: if we should just remove the key from the cache?
+            lru_cache_clone.lock().await.insert(key, 0);
```
ibigbug, line 157: if no proxy is alive maybe we can just remove the entry and build the cache next time?

clash_lib/src/proxy/group/loadbalance/helpers.rs
VendettaReborn: TEST_LRU_STATE is coupled with the strategy fn, so i guess not.
for better readability, the `#[cfg(test)]` lines can be removed, but then there is basically no way to perform the unittest for the lru_cache state inside the closure.

clash_lib/src/proxy/group/loadbalance/helpers.rs
VendettaReborn: i think so, the current solution - use the first entry of proxies as fallback - is written by clash.meta.
it's basically ok, since the cached proxy will be checked before chosen as the return result

ibigbug (approved):

PR #675, opened 2025-01-19
refactor(proxy): make poll_read_exact a general trait for vmess and other streams 

VendettaReborn:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [x] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 💡 Background and solution

there exists some similar code in vmess and shadowtls stream, which i found when i re-read the code. 

the `poll_read_exact` method is especially useful for poll based packet decoding, many proxy projects have their own implement, but there are some ways to make it a general method with a base trait to provide the (inner stream,  BytesMut, read_pos). 
so here comes the base trait `ReadExactBase`, and `ReadExact` trait is automatically implemented for all types which have implemented `ReadExactBase`

```rust
pub trait ReadExactBase {
    /// inner stream to be polled
    type I: AsyncRead + Unpin;
    /// prepare the inner stream, read buffer and read position
    fn decompose(&mut self) -> (&mut Self::I, &mut BytesMut, &mut usize);
}

pub trait ReadExt: ReadExactBase {
    fn poll_read_exact(
        &mut self,
        cx: &mut std::task::Context,
        size: usize,
    ) -> Poll<std::io::Result<()>>;
}
```


clash_lib/src/proxy/group/loadbalance/helpers.rs
```diff
@@ -167,108 +167,3 @@ pub fn strategy_sticky_session(proxy_manager: ProxyManager) -> StrategyFn {
         })
     })
 }
-
-#[cfg(test)]
```
ibigbug, line 171: are these intential?

ibigbug (approved):

VendettaReborn:
need to rebase the master branch. i'll handle it a few days later

PR #680, opened 2025-01-26
refactor(dispatcher): use `RwLock<RunMode>` instead of `Mutex<RunMode>`

cppcoffee:
### 🤔 This is a ...

### 💡 Background and solution

use tokio::sync::RwLock<RunMode> instead of std::sync::Mutex<RunMode>, change dispatch_datagram function to async function.

### DIFF
--- a/clash_lib/src/app/dispatcher/dispatcher_impl.rs
+++ b/clash_lib/src/app/dispatcher/dispatcher_impl.rs
@@ -17,7 +17,7 @@ use std::{
     collections::HashMap,
     fmt::{Debug, Formatter},
     net::SocketAddr,
-    sync::{Arc, Mutex},
+    sync::Arc,
     time::{Duration, Instant},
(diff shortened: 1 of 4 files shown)


hsqStephenZhang (approved):

ibigbug:
why is this better?

> Contrary to popular belief, it is ok and often preferred to use the ordinary [Mutex](https://doc.rust-lang.org/nightly/std/sync/poison/mutex/struct.Mutex.html) from the standard library in asynchronous code.

cppcoffee:
If there are many requests accessing critical section data, using `RwLock` offers better performance; otherwise, `Mutex` performs better.

ibigbug (dismissed):

ibigbug:
https://github.com/Watfaq/clash-rs/pull/680/checks#step:10:236 one more to be updated

PR #689, opened 2025-01-28
style: make poll_read_exact a general trait

VendettaReborn:
re-open of https://github.com/Watfaq/clash-rs/pull/675

### 🤔 This is a ...

- [x] Code style optimization

### 💡 Background and solution

there exists some similar code in vmess and shadowtls stream, which i found when i re-read the code. 

the `poll_read_exact` method is especially useful for poll based packet decoding, many proxy projects have their own implement, but there are some ways to make it a general method with a base trait to provide the (inner stream,  BytesMut, read_pos). 
so here comes the base trait `ReadExactBase`, and `ReadExact` trait is automatically implemented for all types which have implemented `ReadExactBase`

```rust
pub trait ReadExactBase {
    /// inner stream to be polled
    type I: AsyncRead + Unpin;
    /// prepare the inner stream, read buffer and read position
    fn decompose(&mut self) -> (&mut Self::I, &mut BytesMut, &mut usize);
}

pub trait ReadExt: ReadExactBase {
    fn poll_read_exact(
        &mut self,
        cx: &mut std::task::Context,
        size: usize,
    ) -> Poll<std::io::Result<()>>;
}
```

### DIFF
--- a/clash_lib/src/common/io.rs
+++ b/clash_lib/src/common/io.rs
@@ -2,11 +2,13 @@
 use std::future::Future;
 use std::{
     io,
+    mem::MaybeUninit,
     pin::Pin,
     task::{Context, Poll},
     time::Duration,
 };
 
+use bytes::BytesMut;
 use futures::ready;
 use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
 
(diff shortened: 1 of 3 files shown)


ibigbug (approved):
nice!

PR #694, opened 2025-02-09
feat: support ssh protocol 

VendettaReborn:
### 🤔 This is a ...

- [x] New feature

### 💡 Background and solution

tcp connections can be forward by ssh, it's convenient for test, since it needs basically no extra overhead if you own a vps.

russh has full support of ssh protocol, there are two steps for building a proxy connection(to ssh server)

1. auth
2. create tcpip forward channel and convert it into `AnyStream` via russh's client

for authentication, we support password, publickey, keyboard interactive(mostly, totp). the authentication interactions is written according to [rfc425](https://datatracker.ietf.org/doc/html/rfc4252) and [golang crypto library](https://github.com/golang/crypto/blob/9290511cd23ab9813a307b7f2615325e3ca98902/ssh/client_auth.go#L24). 

for tcpip forward channel, the russh's client has provided an ergonomic method, and have implemented `AsyncRead` + `AsyncWrite` for us, we could take advantage of that directly.

i have compared the bin size after this new feature, it increased by 1.5 MB, so i decided to make it an optional feature. 

### example config

configuration format is compatible with [clash.meta's ssh config](https://wiki.metacubex.one/en/config/proxies/ssh/), expect that you could pass 2fa/totp authentication with the provided `totp-opt` field.

```yaml
proxies:
  # Selfhosted
  - name: "ssh-test"
    type: ssh
    server: a.b.c.d
    port: 22
    username: root
    private-key: ~/.ssh/vps1.key
    password: "your password"
    # optional
    totp-opt:
      otp-auth: LGYQ523E66RIT6LWHH6D74XW6Y
```

### 📝 Changelog

support ssh protocol

### TODOs

- [ ] full configuration documentation (perhaps json schema?)

- [x] docker test covers most situations, where 2fa is not enabled
- [x] i have manual tested publickey & password & 2fa on my vps, and it works well, 2fa with totp tutorial can be found here(https://medium.com/@prateek.malhotra004/enhancing-ssh-security-with-two-factor-authentication-2fa-via-pam-and-google-authenticator-70af135c2a95)


ibigbug (approved):
nice

ibigbug:
put `ssh` feature in the ci.yaml build?

PR #708, opened 2025-02-10
ci: add standard and plus feature sets

ibigbug:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [x] Workflow
- [ ] Other (about what?)


Itsusinn:
why test failed on gh action？

ibigbug:
i think test_ssh3 is expected to fail as it's interactive auth?

test_ssh2 I have no idea - @VendettaReborn should this pass ?

clash_lib/src/proxy/ssh/mod.rs
```diff
@@ -450,6 +450,7 @@ mod tests {
     #[cfg(target_os = "linux")]
     #[tokio::test]
     #[serial_test::serial]
+    #[ignore = "failed on github actions"]
```
VendettaReborn: CROSS_CONTAINER_OPTS="--network host" CLASH_DOCKER_TEST="true" cross test --workspace --exclude clash_ffi --target x86_64-unknown-linux-gnu --all-features test_ssh

i have tested the previous fix on my machine, and it seems totaly fine

ibigbug:
```suggestion
```

Issue #405, opened 2024-05-14
Optimize error handling


VendettaReborn:
Motivations: greptimedb has done an excellent job in error handling, see: https://greptime.com/blogs/2024-05-07-error-rust
Shall we adopt some ideas from it?

ibigbug:
yeah the stack trace looks nice. i'd be up for using it

VendettaReborn:
[demo code](https://github.com/VendettaReborn/better-error-experiment), i've written a simple demo, mostly based on greptimedb's impl, since they haven't published the error handling crate yet, it shall be much more convenient to take a glance at their ideas of **virtual user stack** , the blog is already pasted above

ibigbug:
Nice! Planning to make a crate for this?

VendettaReborn:
Nnope, the idea belongs to greptime db, i just wondered how easily can we adopt it in our project. Is this way of error handling better from your point of view? After which i shall start the refactor in this project.

ibigbug:
Ok makes sense. 

Yeah I like it, seems clean way to track error stack and easier to report the where the error originates.

Issue #432, opened 2024-05-27
No address associated with hostname


MrZLeo:
```
thread 'tokio-runtime-worker' panicked at clash_lib/src/app/api/mod.rs:99:76:
called `Result::unwrap()` on an `Err` value: Custom { kind: Uncategorized, error: "failed to lookup address information: No address associated with hostname" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2024-05-27T11:05:15.268235Z ERROR clash_lib/src/lib.rs:158: panic hook: PanicInfo { payload: Any { .. }, message: Some(called `Result::unwrap()` on an `Err` value: Custom { kind: Uncategorized, error: "failed to lookup address information: No address associated with hostname" }), location: Location { file: "clash_lib/src/app/api/mod.rs", line: 99, col: 76 }, can_unwind: true, force_no_backtrace: false }
```

ibigbug:
what is the `external_controller` in your config file?

MrZLeo:
@ibigbug Hi, thinks for reply. It looks like configuration provided by owner doesn't fill in hostname:

```
external-controller: ':9990'
```

But I am wondering whether it is a must? This works fine with old clash.

ibigbug:
i see. i can make a compatibility change for that

Issue #216, opened 2023-12-14
closed with error crypto error: BAD_DECRYPT


xmh0511:
```
grpc initialization error
closed with error crypto error: BAD_DECRYPT
 error: invalid peer certificate: Other(UnsupportedCertVersion)
```

ibigbug:
UnsupportedCertVersion

Likely to be https://github.com/rustls/webpki/issues/29#issuecomment-1453783741

xmh0511:
`clash-rs` should support this certification, the core clash written by Golang supports this configuration and can work well.

xmh0511:
https://github.com/keiko233/clash-nyanpasu/issues/54

ibigbug:
I don't know how to reproduce this, if the issue is the cert, if you don't mind sharing a server that raises the error.

xmh0511:
> I don't know how to reproduce this, if the issue is the cert, if you don't mind sharing a server that raises the error.

You can test it with https://github.com/keiko233/clash-nyanpasu, it has other clash cores, under the same server configuration files, the other core can work but `clash-rs` cannot.

I can supply you with this one server
`{ name: '香港 01 | 专线', type: vmess, server: hk01.84d4d2ae-211c-475f-9076-35fdd96d779b.940a60a7-6cb5-4fa7-bd73-23230d77aa0e.yiyuanjichang.org, port: 443, uuid: 29c79018-29c3-4dd1-aa86-1364f4bd8330, alterId: 0, cipher: auto, udp: true, tls: true, skip-cert-verify: true, servername: www.12306.cn, network: grpc, grpc-opts: { grpc-service-name: '12306' } }`

ibigbug:
Thanks. Will have a look later. I've saved the info, you can delete it if you want.

xmh0511:
> Thanks. Will have a look later. I've saved the info, you can delete it if you want.

Ok. if you want the complete configuration file, I can send it to your email for security.

ibigbug:
No that should be enough. Will let you know if I need more details.

xmh0511:
@ibigbug I tested the latest version, It still doesn't work. The given node can have a timeout test however cannot open any website through the node. Moreover, some other nodes cannot have a timeout test.
![452590302](https://github.com/Watfaq/clash-rs/assets/17973891/fbdd9d1d-f084-4de0-b2ea-6c21971feb04)
[...]

PR #717, opened 2025-02-14
feat: snafu error with better stack trace

VendettaReborn:
### 🤔 This is a ...

- [x] Refactoring
- [x] Code style optimization

### 🔗 Related issue link

https://github.com/Watfaq/clash-rs/issues/405

#### background

greptimedb has done a great job in engineerizing error handling in rust. currently, we are in great debt of error-handlings, it's inconvenient for debugging(i had a deep feeling about that when implementing zero-copy). 

so i tried a POC project in https://github.com/VendettaReborn/better-error-experiment, and it worked as expected, with NOT TOO MUCH overhead added. 

And i just tried it our project, removing `anyhow` in both ClashDnsResolver and Hysteria protocol implementation, i think it may serve as a draft and better POC now. 

#### solution

1. use snafu instead of thiserror/anyhow

```rust
#[derive(Snafu)]
#[snafu(visibility(pub))]
#[stack_trace_debug]
pub enum DnsError {
    #[snafu(display("invalid domain: {domain}"))]
    InvaldDomain {
        domain: String,
        #[snafu(implicit)]
        location: Location,
    },
    #[snafu(display("hickory proto"))]
    Proto {
        #[snafu(implicit)]
        location: Location,
        #[snafu(source)]
        error: hickory_proto::ProtoError,
    },
}
```
2. implement(copied from greptimedb) `stack_trace_debug` macro, which will take the `location` field as the file location in backtrace, and will take `error` field as the external error field. it's important that the two field name, 'location' and 'error' is hardcoded, but that's enough for now.

3. as you can see, with the propagate of Snafu Error, the `Location` is accumulated as a list, which will serve as the virtual error stack at the end. 

4. overhead: **after defining all error kinds properly**, the only overhead is adding `.context(xxxSnafu)` before using `?` or `xxxSnafu.fail()` for rust-style error propagate.
 

### TODOs 

- [ ] discuss if we should adopt this error handling approach (the cost of migrating .etc)
- [ ] move snafu stack trace related code to properer positions instead of `crates/...`
- [ ] register StatusCode for each unique error type
- [ ] migrate legacy code' error handling logic to Snafu, adjust workspace structure

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1222,7 +1222,10 @@ dependencies = [
  "sha2",
  "shadowsocks",
  "smoltcp 0.12.0",
+ "snafu",
  "socket2 0.5.8",
+ "stack-error",
+ "stack-error-macro",
  "tempfile",
  "thiserror 2.0.11",
  "time",
@@ -1728,7 +1731,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "sha3",
- "strum",
+ "strum 0.26.3",
  "syn 2.0.96",
  "void",
 ]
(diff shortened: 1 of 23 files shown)


VendettaReborn:
bad(for developers) error msgs: 

- https://github.com/Watfaq/clash-rs/issues/216
- https://github.com/Watfaq/clash-rs/issues/432

VendettaReborn:
more contexts:

- https://www.reddit.com/r/rust/comments/1cp8xtx/error_handling_for_large_rust_projects/
- https://www.reddit.com/r/rust/comments/1eifu9r/best_practices_for_error_handling_in_big_backend/
- https://www.reddit.com/r/rust/comments/1hn8cqc/error_handling_in_rust_choosing_between_thiserror/

VendettaReborn:
in this error, we can do similar things thanks to `backtrace` field, but it requires `error_generic_member_access` feature of 
 compiler, and the backtrace msg is not user-friendly/readable, code:
https://gist.github.com/VendettaReborn/4876cc32aacb149b93c44bcc13c62f9d

PR #725, opened 2025-02-16
fix: remove android target SO_MARK

Itsusinn:
### 🤔 Bug fix

SO_MARK and SO_BINDTODEVICE are not usable when targting android.


Itsusinn:
#[allow(unused_imports)] is just a workaround. I gonna fix it in another pr

ibigbug (approved):

PR #726, opened 2025-02-16
refactor: conditional compile when glibc static-crt

Itsusinn:
### 🤔 This is a ...

- [x] Refactoring


ibigbug:
why we need this change?

Itsusinn:
1. We should use “real” SystemResolver as much as possible (before changes gnu with -crt-static use hickory for example) 
2. Android dont support `TokioResolver::tokio_from_system_conf` but allow using “real” SystemResolver(libc's getaddrinfo)

Itsusinn:
> i think the `TokioResolver::tokio_from_system_conf` was added for `libc::getaddrinfo` causing crash in certain cases, if we don't have that issue anymore we should look into remove the system TokioResolver some day

`libc::getaddrinfo` will fail under static gnu libc, it a bug existing for decades.
That means only when we have static-crt of gnu targets, we have to keep system TokioResolver

ibigbug:
ok fair engouth. thanks!

ibigbug (dismissed):

Itsusinn:
https://source.android.com/docs/core/ota/modular-system/dns-resolver?hl=zh-cn#dns-q

getaddrinfo is complicated， and broken on cross-test

It works on real android device，tho.

Itsusinn:
From cross-rs

> [1] libc = bionic; Only works with native tests, that is, tests that do not depends on the Android Runtime. For i686 some tests may fails with the error assertion failed: signal(libc::SIGPIPE, libc::SIG_IGN) != libc::SIG_ERR, see https://github.com/cross-rs/cross/issues/140 for more information.

I gonnna skip test for android target

.github/workflows/ci.yml
```diff
@@ -274,32 +275,21 @@ jobs:
...
-          command: test
-          args: --workspace --exclude clash_ffi --target ${{ matrix.target }} ${{ matrix.extra-args }}
-        env:
+          CLASH_DOCKER_TEST: ${{ (startsWith(matrix.os, 'ubuntu') && !matrix.no-docker-test) && 'true' || 'false' }}
           RUSTFLAGS: ${{ matrix.rustflags || '--cfg tokio_unstable' }}
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v5
-        if: ${{ !cancelled() && matrix.coverage }}
+        if: ${{ !cancelled() && matrix.coverage && github.ref == 'refs/heads/master' }}
```
ibigbug: We can upload on PRs to compare cov changes.

ibigbug (approved):

PR #734, opened 2025-02-23
refactor: make shadow tls a general transport method like tls

VendettaReborn:
### 🤔 This is a ...

- [x] Refactoring

### 💡 Background and solution

- make shadow-tls a general transport option(with a few code style optimize), so it can be decoupled with ss protocol. in theory, it can work with protocols of simple authentication like socks5, saving the cost of encryption/decryption. 
- i have tested the combination of socks5 with shadow-tls, and it works well on my machine, but it's better to be submitted in a separate PR.


ibigbug (approved):

ibigbug (approved):

VendettaReborn:
i am still thinking about a better file structure of transport methods(ws, grpc, tls) and obfs methods(simpleobfs, v2ray-ws, shadowtls). do u prefer flatten structure or nested one? since the trasnport & obfs methods are mostly binded to some protocol, it's reasonable that we adopted nested strategy. however, it's kinda messy, with `transport` dir on the same level of other outbound protocols. @ibigbug @Itsusinn thoughts?

- mihomo project structure

image


- sing-box project strcture
- 
image

ibigbug:
i think we can have another trait that does `proxy_stream(in: AnyStream) -> AnyStream` and transports such as grpc/h2 which are standard web protos and others like v2ray/shadow etc, which are non-standard can all impl the trait. and in the outbound, it just compose different transport with proxy_stream(or some other name) with the given ouboudn options.

this will somehow look similar to mihomo concepts of transport and protocol.

ibigbug:
further more I think logically, vmess/shadowsocks/grpc/h2 these kind of things should be same level (transport protocols)

and for h2/grpc these kind of standard protos, we don't need to impl them ourselves. so each folder may only contains the constructors. and for other protocols we need both constructors and actual implementations

and an `outbound` is a composer of different transport protocols

Issue #733, opened 2025-02-23
Is it possible to remove the [patch.crates-io] entry in Cargo.toml


123485k:
I noticed that you made some changes to the source code of `tokio-rustls` and `rustls`, which seems to be not very maintainable. Can you separate these changes into a new crate?

VendettaReborn:
thanks for your concerning.

unfortunately, it's not workable. `shadow-tls` protocol relies on `rustls` with a customized session id generator, which is impossible to introduce in the official crate. [`shadow-tls` official repo](https://github.com/ihciah/shadow-tls/) had done the similar tricks of patching `rustls` and `tokio-rustls`.

it's not a neat solution, i know that, but `shadow-tls` is perhaps one of the best obfs methods based on tls. (another one is reality, maybe). so we deem it a worhty work around. besides, we had kept the patches of `rustls` as neat as possible, and it has been migrated from 0.22 into 0.23 successfully, without **that** much overhead.

Itsusinn:
@VendettaReborn Can we have a copies of these crates?

We rename forked rustls into watfaq-rustls, use rustls & watfaq-rustls at the same time. Then only use watfaq-rustls when necessary.

VendettaReborn:
we can compare the binary size first.

VendettaReborn:
`cargo bloat --release --crates` results:

```txt
 File  .text     Size Crate
 4.3%  37.5%   3.5MiB clash_lib
 1.1%   9.2% 877.5KiB std
 0.4%   3.9% 369.5KiB rustls
 0.3%   3.0% 288.0KiB russh
 0.3%   3.0% 285.9KiB regex_automata
 0.3%   2.7% 253.1KiB quinn_proto
 0.3%   2.6% 251.7KiB hickory_proto
 0.3%   2.3% 219.9KiB axum
 0.3%   2.2% 214.3KiB tokio
 0.3%   2.2% 210.4KiB regex_syntax
 0.2%   2.0% 188.3KiB internal_russh_forked_ssh_key
 0.2%   1.9% 185.6KiB clap_builder
 0.2%   1.7% 166.8KiB h2
 0.2%   1.5% 143.1KiB aho_corasick
 0.1%   1.3% 120.6KiB ring
 0.1%   1.1% 104.1KiB hickory_server
 0.1%   1.0%  97.9KiB quinn
 0.1%   0.8%  80.8KiB [Unknown]
 0.1%   0.7%  65.9KiB hyper
[...]

PR #736, opened 2025-02-24
fix: use watfaq-rustls instead of patched one

VendettaReborn:
### 🤔 This is a ...

- [x] Code style optimization

### 🔗 Related issue link

https://github.com/Watfaq/clash-rs/issues/733

### 💡 Background and solution

- rename rustls into watfaq-rustls, rename tokio-rustls into tokio-watfaq-rustls

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -463,6 +463,31 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
 
 [[package]]
+name = "aws-lc-rs"
+version = "1.12.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4cd755adf9707cf671e31d944a189be3deaaeee11c8bc1d669bb8022ac90fbd0"
+dependencies = [
+ "aws-lc-sys",
+ "paste",
+ "zeroize",
+]
+
+[[package]]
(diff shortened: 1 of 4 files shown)


VendettaReborn:
kinda tricky if we change the package name from rustls into watfaq-rustls. too many direct usage of `rustls::`

Itsusinn:
can this a be small crate？we can use
rustls =  { git =...    package = watfaq-rustls } in its cargo without chaning its imports

VendettaReborn:
> can this a be small crate？we can use rustls = { path =... package = watfaq-rustls } in its cargo.toml without chaning its imports

you mean replace `rustls` with `watfaq-rustls` with this declare of `package = "watfaq-rustls"` but not let them co-exist? this approach won't compile

VendettaReborn:
```bash
$ /usr/bin/ls -al ./target/release/clash-rs # before
-rwxr-xr-x 2 root root 15448224 Feb 24 17:30 ./target/release/clash-rs

$ /usr/bin/ls -al ./target/release/clash-rs # after
-rwxr-xr-x 2 root root 15757672 Feb 24 17:25 ./target/release/clash-rs
```

VendettaReborn:
i know, but since it's only used by `shadow-tls`, i donnot think it necessary to use the alias. better to keep the usage explicit, i think.

ibigbug (approved):
ty! this should help us avoid potential issues upgrading tls and dealing with crates that has rustls as a dependenty

PR #738, opened 2025-02-26
WIP: refactor ss plugin & transport

VendettaReborn:
### 🤔 This is a ...

- [x] Refactoring

### 🔗 Related issue link

https://github.com/Watfaq/clash-rs/pull/734

### 💡 Background and solution

1. move ss plugins into `transport` dir
2. add `sip003::Plugin` as the adapter trait for ss plugins, create `Box<dyn sip003::Plugin>` when initializing Shadowsocks Handler, instead of resolving the config during every call of `connect_stream`.
3. ss v2ray plugin test

### 📝 Changelog

- refactor, generalize transport and shadowsocks plugin logic
- add test for ss v2ray ws plugin

- [x] docs are not needed since its a refactor

@ibigbug @Itsusinn thoughts? does this way of organizing plugins and transport methods better? is there any other advice?


clash_lib/src/proxy/vmess/mod.rs
```diff
@@ -44,8 +35,9 @@ pub struct HandlerOptions {
     pub alter_id: u16,
     pub security: String,
     pub udp: bool,
-    pub transport: Option<VmessTransport>,
-    pub tls: Option<transport::TLSOptions>,
+    pub transport: Option<Box<dyn Transport>>,
```
ibigbug, line 38: it's probably nit: do you think we still keep the Transport options here and only do lazy try_into() in the Handler::new

hsqStephenZhang: makes sense.

hsqStephenZhang: the neatness of the code is also a big concern of this PR

VendettaReborn: makes sense

VendettaReborn: to adjust the previous judgement, i think we could keep the clumsy, dirty part of the code in `converts` dir for cleanliness of the vmess/trojan protocol impl's code. the current code style seems favorable to me.

ibigbug: yeah

ibigbug (approved):
lgtm. The nit change can be done separately.

PR #750, opened 2025-03-12
feat(tun): add default tun device name

lmmqxyx404:
As title described


Itsusinn:
`fd://345` is working on Android.
You might need impl different default values for `device-id: Option<Url>`.

Return Err on android for example

ibigbug:
@lmmqxyx404 does this resolve your concern?

Itsusinn (dismissed):

lmmqxyx404:
OK

ibigbug (approved):

PR #751, opened 2025-03-13
feat: add sentry and panic report

ibigbug:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -4,3 +4,4 @@ rustflags = ["--cfg", "tokio_unstable"]
 
 [env]
 RUST_LOG = { value = "clash=trace" }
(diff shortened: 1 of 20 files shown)


Itsusinn:
Could we remove all panic and replace with Result？

ibigbug:
i think it's still necessary to panic if errors happened that we can't do anything or requires attention

ibigbug:
yes some of them are runtime errors for which there's no where to "return" them to catch

Issue #773, opened 2025-04-13
v0.7.7 crash


tubzby:
Tun mode, fake_ip enabled, working as gateway.

```
Apr 13 14:31:20 raspberrypi clash[7236]: 25-04-13 14:31:20 DEBUG clash_lib/src/app/dispatcher/dispatcher_impl.rs:381: UDP NAT for packet: UdpPacket { src_addr: Domain("ogads-pa.clients6.google.com", 443), dst_addr: Ip(192.168.8.189:62771) }, session: [UDP] 192.168.8.189:62771 -> ogads-pa.clients6.google.com:443
Apr 13 14:31:20 raspberrypi clash[7236]: Well, this is embarrassing.
Apr 13 14:31:20 raspberrypi clash[7236]: clash-rs had a problem and crashed. To help us diagnose the problem you can send us a crash report.
Apr 13 14:31:20 raspberrypi clash[7236]: We have generated a report file at "/tmp/report-7d1657b9-4d98-45e9-88db-18be27cdff91.toml". Submit an issue or email with the subject of "clash-rs Crash Report" and include the report as an attachment.
Apr 13 14:31:20 raspberrypi clash[7236]: - Homepage: https://github.com/watfaq/clash-rs
Apr 13 14:31:20 raspberrypi clash[7236]: - Authors: https://github.com/Watfaq/clash-rs/graphs/contributors
Apr 13 14:31:20 raspberrypi clash[7236]: To submit the crash report:
Apr 13 14:31:20 raspberrypi clash[7236]: Open an issue on GitHub: https://github.com/Watfaq/clash-rs/issues
Apr 13 14:31:20 raspberrypi clash[7236]: We take privacy seriously, and do not perform any automated error collection. In order to improve the software, we rely on people to submit reports.
Apr 13 14:31:20 raspberrypi clash[7236]: Thank you kindly!
```

```
pi@raspberrypi:~ $ cat /tmp/report-7d1657b9-4d98-45e9-88db-18be27cdff91.toml
"name" = "clash-rs"
"operating_system" = "Debian 12.0.0 (bookworm) [64-bit]"
"crate_version" = "0.7.7"
"explanation" = """
Panic occurred in file 'clash_lib/src/session.rs' at line 160
"""
"cause" = 'not a socket address Domain("ogads-pa.clients6.google.com", 443)'
"method" = "Panic"
"backtrace" = """
   0:           0x905e18 - <unresolved>
   1:           0x906718 - <unresolved>
[...]

PR #775, opened 2025-04-13
fix(tun): use session's original dst addr as the src addr during udp NAT

VendettaReborn:
### 🤔 This is a ...

- [x] Bug fix
- [x] Doc

### 🔗 Related issue link

https://github.com/Watfaq/clash-rs/issues/773

### 💡 Background and solution

1. add docs to the rather complicated implementation in `dispatch_datagram`
2. use session's original dst addr as the src addr during NAT when receiving the remote packet(needed only for UDP)

### DIFF
--- a/clash_lib/src/app/dispatcher/dispatcher_impl.rs
+++ b/clash_lib/src/app/dispatcher/dispatcher_impl.rs
@@ -263,6 +263,8 @@ impl Dispatcher {
                     }
                 };
 
+                // for TUN or Tproxy, we need the original destination address
+                let orig_dest = packet.dst_addr.clone();
                 sess.source = packet.src_addr.clone().must_into_socket_addr();
                 sess.destination = dest.clone();
 
@@ -339,7 +341,7 @@ impl Dispatcher {
                             while let Some(packet) = remote_r.next().await {
                                 // NAT
                                 let mut packet = packet;
-                                packet.src_addr = sess.destination.clone();
+                                packet.src_addr = orig_dest.clone();
                                 packet.dst_addr = sess.source.into();
 
                                 debug!(
--- a/clash_lib/src/app/dispatcher/dispatcher_impl.rs
+++ b/clash_lib/src/app/dispatcher/dispatcher_impl.rs
@@ -80,14 +80,14 @@ impl Dispatcher {
         mut sess: Session,
         mut lhs: Box<dyn ClientStream>,
     ) {
-        let dest: SocksAddr = match reverse_lookup(&self.resolver, &sess.destination).await
-        {
-            Some(dest) => dest,
-            None => {
-                warn!("failed to resolve destination {}", sess);
-                return;
-            }
-        };
+        let dest: SocksAddr =
+            match reverse_lookup(&self.resolver, &sess.destination).await {
+                Some(dest) => dest,
+                None => {
+                    warn!("failed to resolve destination {}", sess);
+                    return;
+                }
+            };
 
         sess.destination = dest.clone();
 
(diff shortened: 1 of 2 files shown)


ibigbug:
thanks. I'm going to merge this and https://github.com/Watfaq/clash-rs/compare/master...fix-udp-nat-fakeip together to add some tests for this

ibigbug:
or you can take it over if you don't mind

clash_lib/src/app/dispatcher/dispatcher_impl.rs
```diff
@@ -472,6 +438,50 @@ impl Dispatcher {
     }
 }
 
+// helper function to resolve the destination address
+// if the destination is an IP address, check if it's a fake IP
+// or look for cached IP
+// if the destination is a domain name, don't resolve
+async fn resolve(
```
ibigbug: maybe we call this reverse_resolve or something?

clash_lib/src/app/dispatcher/dispatcher_impl.rs
```diff
@@ -252,57 +226,49 @@ impl Dispatcher {
         let mode = self.mode.clone();
         let manager = self.manager.clone();
 
+        #[rustfmt::skip]
+        /*
+         *  implement details
```
ibigbug, line 231: thanks

clash_lib/src/app/dispatcher/dispatcher_impl.rs
```diff
@@ -252,57 +226,49 @@ impl Dispatcher {
...
+         *    remote_sender, remote_forwarder: channel pair
+         *
+         *  data flow:
+         *    => local_r => init packet => connect_datagram => remote_sender     => remote_forwarder         => remote_w
+         *    => local_w                                    <= remote_receiver_r <= NAT + remote_receiver_w  <= remote_r
+         *  
+         *  notice:
+         *    the NAT is binded to the session in the dispatch_datagram function arg and the closure
+         *    so we need not to add a global NAT table and do the translation
+         *    but the price is, we have to spawn two tasks for each session
```
ibigbug: not quite following this - we need two tasks for bi-directional transfer anyways?

clash_lib/src/app/dispatcher/dispatcher_impl.rs
VendettaReborn: does `reverse_lookup` sounds better?

clash_lib/src/app/dispatcher/dispatcher_impl.rs
VendettaReborn: it's just comments for the current implementation... since it's rather confusing each time reading this function

ibigbug: Yeah what I wasn't sure is you seem to indicate a global nat table would "save the price of spawning two tasks"?

clash_lib/src/app/dispatcher/dispatcher_impl.rs
ibigbug: Yeah sg

clash_lib/src/app/dispatcher/dispatcher_impl.rs
VendettaReborn: oh, sorry, that's not what i mean. i'd better make it clear:
a global nat table is a different approach, which would perhaps save some costs of task spawning, but might introduce downsides of more lock contention. 
personally, i prefer the current one, of course. but i'm not sure the global nat table might be better(it's not full measured, after all)

ibigbug: i see. maybe we could verify that NAT type we support first if we going to change that. I _think_ the current implementation should support full cone, but not very sure.

ibigbug (dismissed):

ibigbug:
plz fmt the code, thanks!

ibigbug (approved):

Issue #778, opened 2025-04-16
'allow_lan' is not propgated to HTTP inbound.


tubzby:
I have already set:

```yaml
bind-address: 0.0.0.0
allow-lan: true
```

But failed with:

```console
Apr 16 09:05:59 gdcx clash[1148134]: 25-04-16 09:05:59  WARN clash_lib/src/proxy/http/inbound/mod.rs:64: Connection from 192.168.x.x:65356 is not allowed
```

And this is inconsistent with `v0.7.6`

ibigbug:
can you try the build here https://github.com/Watfaq/clash-rs/pull/781

tubzby:
@ibigbug confirmed.

ibigbug:
Thanks!

PR #779, opened 2025-04-16
fix: allow lan

Itsusinn:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link

https://github.com/Watfaq/clash-rs/issues/778

### 💡 Background and solution

We didn't respect top level `allow-lan`. This PR deliver the allow-lan to inbound manager.

I thought this can be done in `clash_lib/src/proxy/converters`. But i found I have to copy `InboundOpts` and `CommonInboundOpts`, define new `InboundOptsDef` and `CommonInboundOptsDef` in `clash_lib/src/config/def.rs`

Therefore I just used lazy way

### DIFF
--- a/clash_lib/src/app/inbound/manager.rs
+++ b/clash_lib/src/app/inbound/manager.rs
@@ -136,7 +136,7 @@ impl InboundManager {
                     dispatcher: self.dispatcher.clone(),
                     authenticator: self.authenticator.clone(),
                     listener: inbound.clone(),
-                    allow_lan: self.allow_lan
+                    allow_lan: self.allow_lan,
                 },
             );
         }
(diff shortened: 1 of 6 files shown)


ibigbug:
can we pass in c.allow_lan.unwrawp_or_defaul() here for listeners? https://github.com/watfaq/clash-rs/blob/c8dfdffd3df1f35775b2b57373f43baa0e345ac6/clash_lib/src/config/internal/convert/listener.rs#L96

Itsusinn:
> can we pass in c.allow_lan.unwrawp_or_defaul() here for listeners? https://github.com/watfaq/clash-rs/blob/c8dfdffd3df1f35775b2b57373f43baa0e345ac6/clash_lib/src/config/internal/convert/listener.rs#L96

No too much difference

ibigbug:
That way you don't need to add a new field on General and read it from two places ?

PR #781, opened 2025-04-16
fix(listener): pass allow-lan down to listeners

ibigbug:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link

#778 


Itsusinn:
I think this leads to inconsistent behavior between `clash-rs` and `clash-meta`.
Some users might want one inbound allow lan, meanwhile another disallow lan.
In former implemention, every inbound has its own `allow-lan`.
Thought `clash-meta` use global `allow-lan`, it can use `bind` to forcing listener listen on a non-local ip.

PR #789, opened 2025-05-11
chore: fix shadowquic ci failed

spongebob888:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

udp over datagram of shadowquic use quic udp extension to transport udp. It is lossy and CI may failed 
sometimes due to potential packet loss and should be disabled.

 The correctness should be guaranteed  by upstream.

### DIFF
--- a/clash_lib/src/proxy/shadowquic/mod.rs
+++ b/clash_lib/src/proxy/shadowquic/mod.rs
@@ -288,23 +288,6 @@ mod tests {
 
     #[tokio::test]
     #[serial_test::serial]
-    async fn test_shadowquic_over_datagram() -> anyhow::Result<()> {
-        initialize();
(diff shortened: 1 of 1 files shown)


ibigbug:
Any idea why it's failing ?

spongebob888:
I think it's because the quic datagram extension is inherently unreliable.  So test failed sometimes.
But I don't know why this happens in a localhost setup

ibigbug:
Could you create an issue so we can fix it when we figure it out some day

spongebob888:
sure

PR #808, opened 2025-06-01
refactor(tun): replace cmd style auto route with api on windows

spongebob888:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Enhancement feature
- [x] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 💡 Background and solution
This PR replaces cmd style auto route to windows api
This is done by reimplementing `add_route`

Besides this, it's also found dns hijacking is never working on windows.  This is because tun dns server is not set and
windows won't query dns through tun in such case. To solve this `set_dns_v4/set_dns_v6` is provided to add dns servers.
The default dns server is hard coded to be `1.1.1.1`. This is not a good idea. But we can't set dns server to clash itself  because windows only support dns server on port 53.
If you have better ideas, I'm glad to fix this.

Also, `add_address` api is implemented to provide ipv6 support in the future.

This is experimental and may need more test.

### DIFF
--- a/clash_lib/src/proxy/tun/routes/mod.rs
+++ b/clash_lib/src/proxy/tun/routes/mod.rs
@@ -86,8 +86,10 @@ pub fn maybe_add_routes(cfg: &TunConfig, tun_name: &str) -> std::io::Result<()> 
                     // because it may not be on standard port 53
                     // Windows only support DNS server on port 53
                     let name_server = vec!["1.1.1.1".parse().unwrap()];
-                    let _ = windows::set_dns_v4(&tun_iface, &name_server)
-                        .map_err(|e| tracing::error!("failed to set dns due to:{}", e));
+                    let _ =
+                        windows::set_dns_v4(&tun_iface, &name_server).map_err(|e| {
+                            tracing::error!("failed to set dns due to:{}", e)
+                        });
                 }
                 #[cfg(target_os = "macos")]
                 {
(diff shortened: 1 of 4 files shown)


ibigbug (approved):

clash_lib/src/proxy/tun/routes/mod.rs
```diff
@@ -79,7 +79,18 @@ pub fn maybe_add_routes(cfg: &TunConfig, tun_name: &str) -> std::io::Result<()>
...
                     add_route(&tun_iface, &r)?;
                 }
-
+                #[cfg(target_os = "windows")]
+                {
+                    // Set DNS server or DNS hijack won't work
+                    // We can't set name server to clash DNS listener address
+                    // because it may not be on standard port 53
+                    // Windows only support DNS server on port 53
+                    let name_server = vec!["1.1.1.1".parse().unwrap()];
```
ibigbug, line 88: so when this is set, the system will send DNS requests to the tun device which will then get hijacked, i.e. what's being set here doesn't really matter?

spongebob888: right, if hijacking enabled, any ip is ok

spongebob888:
There is still a bug. Connection get stuck sometimes.

spongebob888:
The bug is due to my personal computer setting. It's not universal. Should work now

PR #809, opened 2025-06-01
feat: Switch to SQLite cache and improve smart proxy group config & state

noctiro:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [x] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 📝 Changelog

- Fixed a possible timing calculation bug in smart proxy groups.

- Shared smart proxy group calculations globally to avoid unnecessary computation and improve performance.

- Migrated cache storage from YAML format to SQLite database (cache.db).

#### 🗂️ SQLite Cache Schema Overview

1.  **`selected` Table**
    *   **Columns:**
        *   `group_name` (TEXT, PRIMARY KEY): The name of the proxy group.
        *   `server` (TEXT): The name of the server currently selected for this group.

2.  **`ip_to_host` Table**
    *   **Columns:**
        *   `ip` (TEXT, PRIMARY KEY): A fake IP address assigned by the Fake-IP DNS feature.
        *   `host` (TEXT): The original hostname corresponding to the fake IP.

3.  **`host_to_ip` Table**
    *   **Columns:**
        *   `host` (TEXT, PRIMARY KEY): An original hostname.
        *   `ip` (TEXT): The fake IP address assigned to this hostname.

4. `smart_penalty` Table
    *   **Purpose**: Stores performance penalty scores for individual proxies within specific "smart" groups (e.g., `smart-1`, `smart-media`). This score is based on the proxy's recent connection performance within that group (like failure counts) and is used to dynamically adjust the proxy's selection priority within that group.
    *   **Scope**: **Group-Specific**. The same proxy can have different penalty scores in different smart groups.
    *   **Columns**:
        *   `group_name` (TEXT): The name of the smart proxy group.
        *   `proxy_name` (TEXT): The name of a specific proxy within the group.
        *   `value` (REAL): The calculated penalty score for this proxy in this group. Higher values usually mean lower preference. This score decays over time and decreases after successful connections.
        *   `last_update_secs` (INTEGER): The Unix timestamp (seconds) when this penalty was last updated.
    *   **Primary Key**: (`group_name`, `proxy_name`)

5. `smart_site_stats` Table
    *   **Purpose**: Stores detailed historical performance statistics for each proxy when accessing specific target sites (domain or IP). This data is used by smart groups to evaluate the health of a proxy for connecting to a specific target (e.g., historical latency, success rate, stability, trend), enabling more granular routing decisions.
    *   **Scope**: **Global/Site-Specific**. This data is not tied to a specific proxy group and reflects the inherent performance of a proxy when connecting to a particular site.
    *   **Columns**:
        *   `proxy_name` (TEXT): The name of the proxy.
        *   `site` (TEXT): The identifier of the target site (usually a domain, but can be an IP).
        *   `delay_history` (BLOB): A list of recent latency measurements (in binary format) for this proxy connecting to this site.
        *   `success_history` (BLOB): A list of recent success/failure results (in binary format) for this proxy connecting to this site.
        *   `last_attempt_secs` (INTEGER): The Unix timestamp (seconds) of the last connection attempt by this proxy to this site.
    *   **Primary Key**: (`proxy_name`, `site`)


noctiro:
[In this commit](https://github.com/Watfaq/clash-rs/pull/809/commits/831b5aee03b3a96497b3eeea2dd1ed871979c58e), I added a forced health check when a proxy is used by either the url-test group or the smart group.

clash_lib/src/app/profile/mod.rs
```diff
@@ -1,229 +1,577 @@
...
+            conn: Arc::new(Mutex::new(conn)),
+            store_selected,
+            selected_cache: tokio::sync::RwLock::new(LruCache::new(cache_cap)),
+            ip_to_host_cache: tokio::sync::RwLock::new(LruCache::new(cache_cap)),
+            host_to_ip_cache: tokio::sync::RwLock::new(LruCache::new(cache_cap)),
+        }
+    }
+
+    /// Ensure all tables exist for backward and forward compatibility.
+    fn migrate_schema(conn: &Connection) {
```
ibigbug: JSON is absolutely a bad idea that I just chose for simplicity. however I'm not sure if SQLite is the best, maybe some other kv db options so we don't need to worry about schema changes and indexes etc?

noctiro: I think it's unnecessary to support multiple database backends, as clash-rs is a local proxy core with no need for complex data storage or queries, and It's rarely used on large-scale infrastructure.

Smart strategy data varies by network and is best stored locally per device. Centralized control, if needed, can be handled via an external API.

noctiro: I just found [redb](https://github.com/cberner/redb), and I'm actually thinking about using it to replace SQLite.

clash_lib/src/proxy/mod.rs
```diff
@@ -174,7 +175,9 @@ pub enum ConnectorType {
 }
 
 #[async_trait]
-pub trait OutboundHandler: Sync + Send + Unpin + DialWithConnector + Debug {
+pub trait OutboundHandler:
+    Any + Sync + Send + Unpin + DialWithConnector + Debug
```
ibigbug, line 179: why we need this `Any`?

ibigbug (commented):
seems the test is failing due to some ss test server not disposed

```
runner@fv-az1786-241:~/work/clash-rs/clash-rs$ sudo lsof -i :10002
COMMAND    PID USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
ssserver 17733 root   12u  IPv4  42314      0t0  TCP *:10002 (LISTEN)
ssserver 17733 root   13u  IPv4  42315      0t0  UDP *:10002
```

clash_lib/Cargo.toml
```diff
@@ -129,7 +133,7 @@ maxminddb = "0.26"
 public-suffix = "0.1"
 murmur3 = "0.5"
 
-arti-client = { version = "0.28", optional = true, default-features = false, features = ["tokio", "rustls", "static-sqlite"] }
+arti-client = { version = "0.28", optional = true, default-features = false, features = ["tokio", "rustls"] }
```
ibigbug:
```suggestion
arti-client = { version = "0.28", optional = true, default-features = false, features = ["tokio", "rustls", "static-sqlite"] }
```

clash_lib/src/app/outbound/manager.rs
```diff
@@ -828,4 +832,82 @@ impl OutboundManager {
...
+                            "Handler identified as Smart but failed to downcast: {}",
+                            handler.name()
+                        );
+                        None
+                    }
+                }
+                OutboundType::UrlTest => {
+                    // Try downcasting to urltest::Handler
+                    if let Some(urltest_handler) =
+                        handler_any.downcast_ref::<urltest::Handler>()
```
ibigbug, line 871: oh i see the cast here

clash_lib/src/app/outbound/manager.rs
```diff
@@ -104,6 +105,10 @@ impl OutboundManager {
         debug!("initializing connectors");
         m.init_handler_connectors().await?;
 
+        // Ensure healthchecks for Smart and UrlTest group dependencies are running
+        // Using a default interval of 600 seconds (10 minutes) for forced checks
+        m.ensure_group_dependencies_healthcheck(600).await?;
```
ibigbug, line 110: do you see in what cases the healthcheck wouldn't run?

ibigbug: i mean the point of having the healthcheck is to have it running in background, if it's sometimes not running, maybe it's a bug we should fix?

ibigbug:
it's a bit weired thought i can't see what change could cause that

clash_lib/src/app/outbound/manager.rs
noctiro: Part of the evaluation logic for the "smart" proxy group is based on health checks. While debugging, I noticed that health checks are not enabled by default. What was the intended logic for enabling health checks in your design? Are they only activated when the user explicitly sets healthcheck in the configuration file?

noctiro: The logic I added is intended to automatically enable health checks when `proxy-provider` or `a proxy` are used in a group of type `smart` or `url-test`, even if the user hasn't explicitly configured healthcheck in the config file.

ibigbug: I see. I think it was controled by the interval https://github.com/watfaq/clash-rs/blob/993dda8e0493b88bbabf6f59ae3ef9114e188b2a/clash_lib/src/app/remote_content_manager/healthcheck.rs#L97

can you see if it's enabled with a positive interval?

ibigbug:
it does look like a genuine failure, i'd suggest separating the PR into smaller ones and see how it failed

noctiro:
> it does look like a genuine failure, i'd suggest separating the PR into smaller ones and see how it failed

OK. But I've been busy with my studies lately, so I'll just make a PR for a few fixes first. I'll work on the rest next week.

ibigbug:
Thanks. I'll have a look too meanwhile.

PR #863, opened 2025-07-16
feat: adding dual stack

spongebob888:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 💡 Background and solution
As have been discussed, this PR is a draft to support dual stack. 
Finally I take the first approach. Namely using `[::]:port` as a dual stack socket.
To achieve this "Ipv6" Addess is added to `BindAdress`. Unfortunately. Binding to ipv6 address forces us to also implement dualstack for socks/http/mixed. 

### 📝 Changelog
1. Adding ipv6 bind address
2. Adding TProxy dualstack
3. The socks outbound associate destination is 0.0.0.0:0, changed to [::]:0 if it's ipv6
4. TProxy UDP sending pack to local is using TProxy as src addr. Create a new udp to send udp packets back.
5. Adding dualstack for http/mixed/ss inbound\
6. Allowlan for TProxy is commented. No trivial way to get local address

- [x] TProxy
- [ ] Redir
- [x] Socks
- [x] http
- [x] mixed 
- [x] shadowsocks (Not tested)

### DIFF
--- a/clash-lib/src/proxy/tproxy/mod.rs
+++ b/clash-lib/src/proxy/tproxy/mod.rs
@@ -11,7 +11,7 @@ use crate::{
 use async_trait::async_trait;
 use std::{
     collections::{HashMap, hash_map},
-    io, mem,
+    io,
     net::SocketAddr,
     os::fd::AsRawFd,
     sync::Arc,
@@ -117,6 +117,10 @@ impl InboundHandlerTrait for TproxyInbound {
             // IPv6 doesn't require this
             socket.set_ip_transparent_v4(true)?;
         }
+        if self.addr.is_ipv6() {
+            // This might not be neccessary
+            set_ip_transparent_v6(&socket)?;
+        }
 
         socket.set_nonblocking(true)?;
         socket.set_broadcast(true)?;
@@ -150,9 +154,14 @@ fn bind_nonlocal_socket(src_addr: SocketAddr) -> io::Result<UdpSocket> {
         socket2::Domain::IPV6
     };
     let socket = socket2::Socket::new(domain, socket2::Type::DGRAM, None)?;
+    // This is required to allow binding to nonlocal address
     if src_addr.is_ipv4() {
         socket.set_ip_transparent_v4(true)?;
+    } else {
+        set_ip_transparent_v6(&socket)?;
     }
+    socket.set_nonblocking(true)?;
+    socket.set_reuse_address(true)?;
     socket.bind(&src_addr.into())?;
 
     let socket = UdpSocket::from_std(socket.into())?;
@@ -241,6 +250,7 @@ async fn handle_inbound_datagram(
                 }
             }
         }
+        warn!("tproxy udp listening ended");
 
         closer.send(0).ok();
     });
@@ -253,8 +263,6 @@ fn set_ip_recv_orig_dstaddr(
     level: libc::c_int,
     socket: &socket2::Socket,
 ) -> io::Result<()> {
-    let fd = socket.as_raw_fd();
-
     let opt = match level {
         libc::IPPROTO_IP => libc::IP_RECVORIGDSTADDR,
         libc::IPPROTO_IPV6 => libc::IPV6_RECVORIGDSTADDR,
@@ -262,22 +270,7 @@ fn set_ip_recv_orig_dstaddr(
     };
 
     let enable: libc::c_int = 1;
-
-    unsafe {
-        let ret = libc::setsockopt(
-            fd,
-            level,
-            opt,
-            &enable as *const _ as *const _,
-            mem::size_of_val(&enable) as libc::socklen_t,
-        );
-
-        if ret != 0 {
-            return Err(io::Error::last_os_error());
-        }
-    }
-
-    Ok(())
+    set_socket_option(socket, level, opt, enable)
 }
 
 async fn handle_packet_from_dispatcher(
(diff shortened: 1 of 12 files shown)


clash-lib/src/proxy/tproxy/mod.rs
```diff
@@ -68,16 +73,17 @@ impl InboundHandlerTrait for TproxyInbound {
 
         loop {
             let (socket, _) = listener.accept().await?;
-            let src_addr = socket.peer_addr()?;
-            if !self.allow_lan && src_addr.ip() != socket.local_addr()?.ip() {
+            let src_addr = socket.peer_addr()?.to_canonical();
+            // for dualstack socket src_addr may be ipv4 or ipv6
+            if !self.allow_lan && !src_addr.ip().is_loopback() {
```
ibigbug: so this will essentially only allow connection from 127.1

and connecting with 10.0.0.100 to a inbound on 0.0.0.0  from local machine will fail?

clash-lib/src/proxy/tproxy/mod.rs
```diff
@@ -54,12 +58,13 @@ impl InboundHandlerTrait for TproxyInbound {
...
-            socket2::Type::STREAM,
-            None,
-        )?;
-        socket.set_ip_transparent_v4(true)?;
+        let (socket, dualstack) =
+            try_create_dualstack_socket(self.addr, socket2::Type::STREAM)?;
+        set_ip_transparent(&socket, self.addr.is_ipv6())?;
+        if dualstack {
+            // set ipv4 transparent
+            set_ip_transparent(&socket, false)?;
```
ibigbug: this function parameter is confusing.. i think you mean enable for non ipv6 here but passing a false value sounds like disable it...

maybe rename the `set_ip_transparent` to `set_ip_transparent_v6` and stay with using `socket.set_ip_transparent_v4` here if that works?

clash-lib/src/proxy/utils/socket_helpers.rs
```diff
@@ -198,3 +198,42 @@ pub async fn family_hint_for_session(
...
+    tcp_or_udp: socket2::Type,
+) -> std::io::Result<(socket2::Socket, bool)> {
+    let domain = if addr.is_ipv4() {
+        socket2::Domain::IPV4
+    } else {
+        socket2::Domain::IPV6
+    };
+    let mut dualstack = false;
+    let socket = socket2::Socket::new(domain, tcp_or_udp, None)?;
+    if addr.is_ipv6() && addr.ip().is_unspecified() {
```
ibigbug, line 225: why checking `addr.ip().is_unspecified()` here?

spongebob888:
The allow-lan logic is not correct. We can't get local_addr in tproxy since local_addr is actually proxy destination.

clash-lib/src/proxy/tproxy/mod.rs
spongebob888: Yes. I tested this it barely works. Nearly every connection has a non-localhost address even it is from my local process. So this is a problem.

clash-lib/src/proxy/tproxy/mod.rs
spongebob888: Ok

clash-lib/src/proxy/utils/socket_helpers.rs
spongebob888: Dualstack only works when binding to a ipv6 unspecified address

clash-lib/src/proxy/tproxy/mod.rs
spongebob888: Any ideas to implement this?

clash-lib/src/proxy/utils/socket_helpers.rs
ibigbug: ah i see, thanks!

clash-lib/src/proxy/tproxy/mod.rs
ibigbug: i think the original implementation should still work for v6?

```
let listener = std::net::TcpListener::bind("[::]:12345").unwrap();
...

  println!(
                    "New connection from {} -> {}",
                    socket.peer_addr().unwrap(),
                    socket.local_addr().unwrap()
                );
```



```
Server listening on port 12345
New connection from [::ffff:10.0.15.177]:54306 -> [::ffff:10.0.15.177]:12345
Received 0 bytes: []
New connection from [::1]:54308 -> [::1]:12345
Received 0 bytes: []
New connection from [2401:redacted]:54327 -> [2401:redacted]:12345
Received 0 bytes: []
```

ibigbug: but this is for normal socket, not very sure about tproxy case.

> Nearly every connection has a non-localhost address

what these srcs look like?

spongebob888: it is the address of my nic card like 192.168.1.1

ibigbug: right then i think it should work same way by checking incoming.src == socket.local_addr

spongebob888: I'll check that

spongebob888: it doesn't work the socket.local_addr is "[::]:port"

spongebob888:
After some investigation, IPV6 doesn't require seting IP_TRANSPARENT. As shown in https://docs.kernel.org/networking/tproxy.html
> Because of certain restrictions in the IPv4 routing output code you’ll have to modify your application to allow it to send datagrams _from_ non-local IP addresses. All you have to do is enable the (SOL_IP, IP_TRANSPARENT) socket option before calling bind:

spongebob888:
This also solves the puzzle why socket2 doesn't provide `set_ip_transparent_v6`

spongebob888:
UDP seems broken

spongebob888:
all work is done

ibigbug (approved):
thank you

PR #874, opened 2025-07-21
fix: ipv6 not effective for redir/tproxy

spongebob888:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 💡 Background and solution
IPV6 switch is not effective is bindaddress is not explicitly set to "::"

This PR sets the bindaddress to "::" if it is not explicitly set and ipv6 option is on

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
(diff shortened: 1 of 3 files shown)


ibigbug:
if you change the default value here to unspecified and match it with the ipv6 switch in the listener, would it work? https://github.com/watfaq/clash-rs/blob/master/clash-lib/src/config/internal/config.rs#L126

spongebob888:
Then. If user set explicitly `bind-address:0.0.0.0` and `ipv6: true` will change `bind-address:[::]` which may be confusing

spongebob888:
We can't distinguish between user set bind-address to 0.0.0.0 or empty field

ibigbug:
Which should be fine? It they want 0.0.0.0 and ipv6 it would work too if you listen ::? If it's dual stack. 

I just think making it optional adds extra complexity to match it instead of simply unwrap it

spongebob888:
OK

PR #903, opened 2025-08-16
Add VLESS protocol support

ibigbug:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link

N/A - Adding support for VLESS protocol

### 💡 Background and solution

**Problem:**
ClashRS currently supports multiple proxy protocols (Shadowsocks, Trojan, Vmess, Wireguard, etc.) but lacks support for the VLESS protocol, which is a lightweight protocol used in V2Ray ecosystem.

**Solution:**
This PR implements full VLESS protocol support including:
1. Configuration parsing for VLESS proxies in YAML config
2. VLESS protocol handler with UUID-based authentication
3. Transport layer support (WebSocket, gRPC, H2)
4. TLS encryption support
5. UDP relay capability

**Sample Configuration:**
```yaml
proxies:
  - name: "vless-ws-tls"
    type: vless
    server: 127.0.0.1
    port: 8443
    tls: true
    network: ws
    uuid: b831381d-6324-4d53-ad4f-8cda48b30811
    ws-opts:
      path: /websocket
```

### 📝 Changelog

**New Features:**
- Add VLESS protocol support with WebSocket, gRPC, and H2 transport options
- Support TLS encryption for VLESS connections
- Add UDP relay capability for VLESS proxies
- Include comprehensive test configurations

**Breaking Changes:**
None - this is purely additive functionality

**Configuration Changes:**
- New `vless` proxy type available in configuration
- New configuration options: `uuid`, transport-specific options

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>


ibigbug:
@Itsusinn FYI

Itsusinn:
Users usually don't use v2ray's vless.

Instead they use https://github.com/XTLS/Xray-core

Xray-core introduced xtls,vision.
https://github.com/XTLS/Xray-core/discussions/1295
which i think was too complicated to implement 

@ibigbug

ibigbug:
Yes I'm aware of that but all of that requires the plain VLESS implementation to begin with, and we can see if we want to add more of those complicated transports.

Itsusinn:
> Yes I'm aware of that but all of that requires the plain VLESS implementation to begin with, and we can see if we want to add more of those complicated transports. 

I think “XTLS” --- TLS in TLS (and tls-sniffer?) is pretty interesting , could be a common utility

PR #913, opened 2025-08-26
ci: mips build jobs

Itsusinn:
CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+defaults:
+  run:
+    shell: bash
+
 env:
   PACKAGE: "clash"
   REGISTRY: "ghcr.io"
@@ -65,12 +69,6 @@ jobs:
             release-name: x86_64-unknown-linux-gnu-static-crt
             extra-args: -F "plus"
             rustflags: "-Ctarget-feature=+crt-static --cfg tokio_unstable"
-          - os: ubuntu-latest
-            target: i686-unknown-linux-gnu
-            release-name: i686-unknown-linux-gnu-static-crt
-            tool: cross
-            extra-args: -F "plus"
-            rustflags: "-Ctarget-feature=+crt-static --cfg tokio_unstable"
           # Linux arm gnu
           - os: ubuntu-latest
             target: aarch64-unknown-linux-gnu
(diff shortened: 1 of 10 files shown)


Itsusinn:
block on https://github.com/taiki-e/portable-atomic/issues/2

clash-dns/src/handler.rs
```diff
@@ -378,10 +378,26 @@ mod tests {
...
+                .await
+            {
+                Ok(v) => {
+                    response = v;
+                    break;
+                }
+                Err(e) => {
+                    retries -= 1;
+                    if retries == 0 {
+                        panic!("Send DNS query error: {e}");
```
ibigbug: Let's not panic here. Just return the error ?

Itsusinn: this funtion only used in test, so there should be no difference

Itsusinn: But actually all tests should return a Result<T,E>

Itsusinn:
Should we keep mips CI in this form, or we annotate it? @ibigbug 
Currently mips have `continue-on-error`

.github/workflows/ci.yml
```diff
@@ -11,6 +11,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
+defaults:
+  run:
+    shell: bash
```
ibigbug, line 16: Why do we need this

.github/workflows/ci.yml
```diff
@@ -65,12 +70,6 @@ jobs:
             release-name: x86_64-unknown-linux-gnu-static-crt
             extra-args: -F "plus"
             rustflags: "-Ctarget-feature=+crt-static --cfg tokio_unstable"
-          - os: ubuntu-latest
-            target: i686-unknown-linux-gnu
-            release-name: i686-unknown-linux-gnu-static-crt
```
ibigbug, line 70: No more static crt ?

.github/workflows/ci.yml
Itsusinn: For using bash in windows
https://github.com/Watfaq/clash-rs/pull/913/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL255-L260

.github/workflows/ci.yml
Itsusinn: I removed i686 linux gnu static crt because linux i686 gnu have zigcc=2.17 which i think is low enough for all distros

For some other targets having zigcc=2.17, I removed static-crt as well

ibigbug:
Please fix the build

Itsusinn:
> Please fix the build 

The whole continue-on-error mechanism is weired

ibigbug:
Why mips build are showing errors?

PR #923, opened 2025-09-06
feat(dns): adding dns over proxy

spongebob888:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 💡 Background and solution
This PR adds dns over proxy
Only plain proxy supported right now

Tests need to be added


spongebob888:
Using plain outbounds only may be very inconvenient. Maybe we could add proxy groups?

ibigbug:
Proxy groups should work too. We just need to refactor the outbound manager a bit to break the potential circular dependency I think.

spongebob888:
This could be added later since it also involves db downloader

ibigbug:
could you mark this PR ready for review

spongebob888:
Sure

PR #931, opened 2025-09-23
.

arkxfly:


ibigbug:
I think there's  default_bool_true somewhere already

clash-lib/src/config/internal/listener.rs
```diff
@@ -99,3 +102,4 @@ pub struct CommonInboundOpts {
     /// Linux routing mark
     pub fw_mark: Option<u32>,
 }
+
```
ibigbug:
```suggestion
```

PR #970, opened 2025-12-10
Fix get default outbound interface on iOS

hiFrankYoung:


ibigbug:
Please provide a PR comment explaining the new logic and fix the cargo fmt issue. Thanks.

Issue #974, opened 2025-12-24
TProxy bind nonlocal address may collide with other process


spongebob888:
To build a tproxy, we will create a udp listener with ip_transparency option set. And we send the packet received locally to remote. When we received remote packet we send the packet back to local client. However we must bind to nonlocal address the same as remote server address which set the src address of IP packet to remote address. And this nonlocal address port may collide with the port of other process. I mean another process has used this port. This will lead to a address already in use error.

For example: 
Process 1: listening on 0.0.0.0:53

TProxy Process 2 listening on 8900:  suppose we received a udp request to 1.1.1.1:53.

UPLOAD traffic:
``` 127.0.0.1:12345 -> TProxy(0.0.0.0:8900) -> 1.1.1.1:53```

DOWNLOAD traffic:
``` 127.0.0.1:12345 <- clash-rs( create a new udp socket bound to 1.1.1.1:53 )  <- 1.1.1.1:53```

However this creates new udp socket bound to 1.1.1.1:53 will collide with Process 1

spongebob888:
I tried reuse_port like shadowsocks-rust. But it doesn't solve. Since this requires both process set reuse_port sockoption.

ibigbug:
Why is 1.1.1.1:53 and 0.0.0.0:53 considered conflict?

I mean tproxy is quit mature I wouldn't expect this kind of flaw exist unless we didn't set it up properly? Which might be the case so few documentations about it.

spongebob888:
In my laptop. It did conflict. But if replace 0.0.0.0 with address like 127.0.0.1, then no conflict. I will test singbox later.

spongebob888:
Well, I successfully reproduced on sing-box tproxy.

Process 1: a dns server listening on udp 53 port
Process 2: sing-box tproxy inbound listening on 8900

A request to 114.114.114.114:53 was proxied  by sing-box  and it reports an *address already in use error*

spongebob888:
https://github.com/SagerNet/sing-box/issues/3646

spongebob888:
gpt suggests me using sendmsg with unbound socket. I'd give a try

PR #975, opened 2025-12-27
fix(tproxy): using raw socket to sending packet to local

spongebob888:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link
#974 

### 💡 Background and solution
As stated in #974 , Using `bind nonlocal socket` may collide with other process. This PR fixes this by using raw socket 
to sending udp packet back to local clients. 

This requires to construct a UDP packet  manually. So etherparse package is used to build IP header and UDP header.

### DIFF
--- a/clash-lib/src/proxy/tproxy/mod.rs
+++ b/clash-lib/src/proxy/tproxy/mod.rs
@@ -1,6 +1,7 @@
 use super::{inbound::InboundHandlerTrait, tun::TunDatagram};
 use crate::{
     app::dispatcher::Dispatcher,
+    common::errors::new_io_error,
     proxy::{
         datagram::UdpPacket,
         utils::{ToCanonical, apply_tcp_options, try_create_dualstack_socket},
@@ -9,17 +10,11 @@ use crate::{
 };
 
 use async_trait::async_trait;
-use std::{
-    collections::{HashMap, hash_map},
-    io,
-    net::SocketAddr,
-    os::fd::AsRawFd,
-    sync::Arc,
-    time::Duration,
-};
-use tokio::{net::TcpListener, time::Instant};
+use etherparse::PacketBuilder;
+use futures::future;
+use std::{io, net::SocketAddr, os::fd::AsRawFd, sync::Arc, task::Poll};
+use tokio::net::TcpListener;
 use tracing::{trace, warn};
-use unix_udp_sock::UdpSocket;
 
 pub struct TproxyInbound {
     addr: SocketAddr,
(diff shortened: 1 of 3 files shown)


clash-lib/src/proxy/tproxy/mod.rs
```diff
@@ -149,26 +144,92 @@ impl InboundHandlerTrait for TproxyInbound {
...
+            Some(libc::IPPROTO_RAW.into()),
+        )?
     };
-    let socket = socket2::Socket::new(domain, socket2::Type::DGRAM, None)?;
-    // This is required to allow binding to nonlocal address
-    if src_addr.is_ipv4() {
-        socket.set_ip_transparent_v4(true)?;
-    } else {
-        set_ip_transparent_v6(&socket)?;
+    // socket.set_nonblocking(true)?;
```
ibigbug: why this line commented? i think we need unblocking socket for tokio other wise it's quite slow

clash-lib/src/proxy/tproxy/mod.rs
```diff
@@ -282,55 +343,39 @@ fn set_ip_recv_orig_dstaddr(
...
                 // remote -> local
-                let now = Instant::now();
-                let src_addr = pkt.src_addr.must_into_socket_addr();
-                let responder = match responder_map.entry(src_addr) {
-                    hash_map::Entry::Occupied(mut occupied_entry) => {
-                        occupied_entry.get_mut().1 = now;
-                        occupied_entry.get().0.clone()
+                let src_addr = pkt.src_addr.try_into_socket_addr().unwrap();
+                let dst_addr = pkt.dst_addr.try_into_socket_addr().unwrap();
+                match (src_addr, &socket_v4, &socket_v6) {
```
ibigbug, line 361: do you know how would this affect NAT behavior

clash-lib/src/proxy/tproxy/mod.rs
```diff
@@ -149,26 +144,92 @@ impl InboundHandlerTrait for TproxyInbound {
...
+            packet = Vec::<u8>::with_capacity(builder.size(buf.len()));
+        }
+        _ => {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "source and destination address families do not match",
+            ));
+        }
+    }
+    builder
```
ibigbug, line 198: is this expensive to build raw packets everytime?

clash-lib/src/proxy/tproxy/mod.rs
spongebob888: Indeed. I guess I commented by mistake.

clash-lib/src/proxy/tproxy/mod.rs
spongebob888: I don't think this will affect NAT. Since at this stage NAT have been done by dispatcher. This is the final stage to send data back to local clients.

clash-lib/src/proxy/tproxy/mod.rs
spongebob888: I think is's OK. Generally, it is build by ths OS. We just build it ourselves. This works like tun where UDP packet is also built by smoltcp stack. Besides, UDP packet is stateless, the header is pretty simple.

ibigbug (approved):

ibigbug:
@VendettaReborn might want to double check as well

PR #988, opened 2026-01-11
fix(tun): address not set on linux

spongebob888:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 💡 Background and solution
IP address is not set for tun on linux. Is this intended? 
But I think enabling such setting is helpful

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -190,7 +190,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           submodules: true
-      - uses: actions/cache@v4
+      - uses: actions/cache@v5
         with:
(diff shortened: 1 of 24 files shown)


ibigbug:
Yeah it was intentional. We never had an address for the tin until I noticed it's required on Windows to work.

ibigbug:
But I guess it doesn't hurt to add

ibigbug (approved):

PR #1025, opened 2026-02-10
refactor: unify runners API and utilize Cancellation token 

Itsusinn:
Using clash-rs as a library requires complete life cycle management of all tasks spawned by tokio::task::spawn.

I had a idea.

Define a modified `tokio::spawn`

```rust
#[inline(always)]
pub fn spawn<F, Fut>(token: CancellationToken, future: F) -> JoinHandle<()>
where
    F: FnOnce(CancellationToken) -> Fut + Send + 'static,
    Fut: Future + Send + 'static,
    Fut::Output: Send + 'static,
{
    let fut = future(token.child_token());
    tokio_spawn(async move {
        tokio::select! {
            _ = token.cancelled() => {
                // Task was cancelled, return early
                // We can do some cleanup here if needed before
                return;
            }
            output = fut => {
                // future completed, do something with the output if needed
                // I perfer to use a Result or ControlFlow enum here to indicate success or failure instead of panicking
                // Such as: ControlFlow::Continue(()) or ControlFlow::Break(error)
                // We can cancel ourself if it's an error, or just log it and return if it's a success
                // if output.is_break() {
                    // token.cancel();
                // }

                _ = output; // Ignore the output for now
            }
        }
    })
}

```

An example usage

```rust


pub async fn how_to_use(){
    let token = CancellationToken::new();
    spawn(token.child_token(), long_time_polling);

    spawn(token.child_token(), async move |token| {
        _ = token; // Use the token if needed, or just ignore it if not
    
        // Simulate some work in the background task
        tokio::time::sleep(std::time::Duration::from_secs(3)).await;
    }).await.unwrap();

    // Simulate some work in the main task
    tokio::time::sleep(std::time::Duration::from_secs(5)).await;
    // Signal tasks to shutdown
    token.cancel();
}

pub async fn long_time_polling(token: CancellationToken) {
    struct StateMaybeIO;
    impl StateMaybeIO {
        async fn close(&mut self) {
            // Simulate some cleanup work here
            // For TUIC we need send `FIN` to the peer before close the connection, so we can do that here
            // Or close all UdpSesions
            tokio::time::sleep(std::time::Duration::from_secs(2)).await;
        }
    }
    // Simulate a long-running task that checks for cancellation

    let mut state = StateMaybeIO;
    loop {
        tokio::select! {
            _ = token.cancelled() => {
                // Task was cancelled, perform cleanup and return
                state.close().await;
                break;
            }
            _ = tokio::time::sleep(std::time::Duration::from_secs(1)) => {
                // Do some work here...
            }
        }
    }
}

```

### DIFF
--- a/clash-lib/src/app/api/runner.rs
+++ b/clash-lib/src/app/api/runner.rs
@@ -126,7 +126,6 @@ impl Runner for ApiRunner {
         });
         let cancellation_token = self.cancellation_token.clone();
         tokio::spawn(async move {
-            info!("Starting API server");
             let mut router = Router::new()
                 .route("/", get(handlers::hello::handle))
                 .route("/logs", get(handlers::log::handle))
@@ -194,8 +193,11 @@ impl Runner for ApiRunner {
                 };
                 let router_clone = router.clone();
                 Some(async move {
-                    info!("Starting API server on TCP address {bind_addr}");
                     let listener = tokio::net::TcpListener::bind(&bind_addr).await?;
+                    info!(
+                        "API server is listening on TCP address {}",
+                        listener.local_addr()?
+                    );
                     // TCP related security checks
                     if let Ok(addr) = listener.local_addr() {
                         if !addr.ip().is_loopback()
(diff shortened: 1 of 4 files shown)


ibigbug:
@claude[agent] what's left for this PR?

PR #1033, opened 2026-02-25
fix(tun): wait for tun visibility before adding routes

### LINKED ISSUE #1032: Intermittent startup crash in TUN mode: "tun interface not found" (SIGABRT/systemd restart loop)
### Environment
  - OS: Debian 12 (bookworm), Linux
  - clash-rs: v0.9.3
  - Run mode: systemd service
  - TUN enabled (`tun.enable: true`), reproducible with `route-all: true`

  ### What happens
  When starting clash-rs in TUN mode, startup can intermittently fail and enter a systemd restart loop.

  Observed behavior:
  - process exits with `status=6/ABRT`
  - journal shows crash report generation
  - panic message includes: `tun interface not found`

  ### Key logs
  From `journalctl -u net-proxy.service`:
  - `tun device tun0 does not exist, creating.`
  - `setting up routes for tun tun0`
  - then crash (`status=6/ABRT`) and auto-restart

  From crash report (`/tmp/report-*.toml`):
  - `Panic occurred in file 'clash-lib/src/proxy/tun/routes/mod.rs' at line 42`
  - `cause = "tun interface not found"`

  ### Impact
  Service is unstable in TUN mode on startup and may fail to come up reliably under systemd.

kkkiio:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Enhancement feature
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link

fix #1032 

### 💡 Background and solution

On Linux/systemd, TUN startup can fail intermittently with:

- `tun interface not found`

This happens when route setup runs right after `build_async()`, but the new TUN interface is not yet visible from `NetworkInterface::show()`.

This PR adds a short wait before route setup for newly created TUN interfaces:
- retry up to 40 times, 50ms each (max 2s)
- return a clear error if still not visible

### 📝 Changelog

- Fix intermittent TUN startup race on Linux.
- Wait for TUN interface visibility before `maybe_add_routes()`.

### DIFF
--- a/clash-lib/src/proxy/tun/inbound.rs
+++ b/clash-lib/src/proxy/tun/inbound.rs
@@ -10,7 +10,7 @@ use crate::{
     },
 };
 use futures::{SinkExt, StreamExt};
-use std::sync::Arc;
+use std::{sync::Arc, thread::sleep, time::Duration};
 use tracing::{debug, error, info, trace, warn};
 use url::Url;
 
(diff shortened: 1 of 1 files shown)


Itsusinn:
@ibigbug I think we should use async here, std::thread::sleep will break tokio runtime

PR #1057, opened 2026-03-16
feature: support vless reality

SKTT1Ryze:
## Summary

Implements the `xtls-rprx-vision` flow for VLESS proxies running over the Reality transport. Prior to this change, configuring `flow: xtls-rprx-vision` had no effect — the flow field was parsed but never used, causing handshake failures when connecting to Xray servers that require Vision framing.

## Background

VLESS+Reality+Vision involves two distinct mechanisms:

**Vision framing** wraps inner TLS handshake records inside a custom frame format:

```
[UUID: 16 bytes — first frame only]
[command: u8]          0x00 continue | 0x01 end | 0x02 direct (splice)
[content_len: u16 BE]
[padding_len: u16 BE]
[content: content_len bytes]
[padding: padding_len bytes — random, discarded by receiver]
```

**XTLS-splice** is triggered when either side sends `CMD_PADDING_DIRECT` (`0x02`). Both peers then bypass the outer Reality TLS layer and communicate over raw TCP, with only the inner TLS providing encryption. On the Xray server side this is done via `UnwrapRawConn()` which strips the Reality connection down to the underlying `net.Conn`.

The root bug: after `CMD_DIRECT` was exchanged, the Xray server wrote raw inner-TLS bytes directly to TCP, but our `rustls` client was still attempting to decrypt them as Reality TLS records:

```
rustls::DecryptError: cannot decrypt peer's message
```

## Layer stack

```
VisionStream                   owns VisionOptions (writes flags)
  └─ VlessStream               VLESS framing + addon encoding
      └─ SplicableTlsStream    reads flags; bypasses Reality TLS when set
          └─ Reality TLS
              └─ TCP
```

## Handshake sequence

```
Client                                      Xray server
  |                                               |
  |---------- Reality TLS handshake ------------>|
  |<--------- Reality TLS handshake -------------|
  |                                               |
  |========= Vision framing mode ================|
  |--[UUID][CMD=0x00][inner TLS ClientHello]----->|
  |<-[UUID][CMD=0x00][inner TLS ServerHello]------|
  |<-[CMD=0x02][inner TLS AppData]----------------|  server triggers splice
  |--[CMD=0x02][inner TLS AppData]--------------->|  client triggers splice
  |                                               |
  |========= Splice mode (raw TCP) ==============|
  |--[raw inner-TLS AppData]--------------------->|  no outer TLS encryption
  |<-[raw inner-TLS AppData]----------------------|
```

### `transport/splice_tls.rs` (new)

**`VisionOptions`** — a pair of `Arc<AtomicBool>` splice flags shared between the transport layer and the Vision framing layer:

```rust
pub struct VisionOptions {
    pub read_flag:  Arc<AtomicBool>,
    pub write_flag: Arc<AtomicBool>,
}
```

**`SplicableTlsStream`** — wraps a `TlsStream<AnyStream>` (Reality TLS) and monitors the splice flags. When a flag is set, it bypasses `rustls` and reads/writes directly on the underlying TCP socket via `tls.get_mut().0`. On the read side, any plaintext already buffered inside the TLS session is drained via `conn.reader()` before switching to raw mode.

### `transport/reality.rs`

`RealityClient` now overrides `proxy_stream_spliced` on the `Transport` trait. Instead of boxing the TLS stream opaquely, it:

1. Creates a `read_flag` / `write_flag` pair
2. Establishes the Reality TLS connection as a concrete `TlsStream<AnyStream>`
3. Wraps it in a `SplicableTlsStream` bound to those flags
4. Returns `(AnyStream, Some(VisionOptions))`

### `transport/mod.rs`

`Transport` gains a `proxy_stream_spliced` method with a default implementation that delegates to `proxy_stream` and returns `None` for `VisionOptions`. Only `RealityClient` overrides it; all other transports are unaffected.

```rust
async fn proxy_stream_spliced(
    &self,
    stream: AnyStream,
) -> io::Result<(AnyStream, Option<VisionOptions>)> {
    Ok((self.proxy_stream(stream).await?, None))
}
```

### `vless/vision.rs` (new)

`VisionStream` implements `AsyncRead + AsyncWrite` and handles Vision framing end-to-end.

- **Write path**: wraps each outbound buffer in a Vision frame. The first frame prepends the 16-byte client UUID and includes random padding for traffic-analysis resistance. When the payload is a TLS `ApplicationData` record (`0x17`), `CMD_PADDING_DIRECT` is used and `write_flag` is signalled — subsequent writes go raw.
- **Read path**: buffers raw bytes from the inner stream, decodes Vision frames (skipping padding and the server UUID prefix), and delivers content to the caller. On `CMD_PADDING_DIRECT` or `CMD_PADDING_END`, `read_flag` is signalled and remaining bytes pass through raw.

```rust
// Pass Some(VisionOptions) for Reality (splice enabled), None for plain TLS.
pub fn new(inner: AnyStream, uuid: String, opts: Option<VisionOptions>) -> io::Result<Self>
```

### `vless/stream.rs`

`VlessStream` now encodes the `flow` field into the VLESS request header's addon bytes using protobuf field encoding (`0x0A` + varint length + UTF-8 flow string), as required by the Xray VLESS spec. Previously the addon length was always written as `0`.

### `vless/mod.rs` and `converters/vless.rs`

`inner_proxy_stream` is simplified to a single linear flow with no Reality-specific branching:

```rust
let (s, vision_opts) = if let Some(tls) = self.opts.tls.as_ref() {
    tls.proxy_stream_spliced(s).await?
} else {
    (s, None)
};
// ...
if flow == "xtls-rprx-vision" {
    VisionStream::new(Box::new(vless_stream), uuid, vision_opts)
}
```

`HandlerOptions` no longer carries a redundant `reality_client: Option<Arc<RealityClient>>` field.

## Testing

Unit tests cover Vision framing (`vless/vision.rs`) and VLESS addon encoding (`vless/stream.rs`). End-to-end verification against a live Xray server:

```yaml
- name: vless-reality-vision
  type: vless
  flow: xtls-rprx-vision
  tls: true
  reality-opts:
    public-key: <ed25519-public-key>
    short-id: <hex-short-id>
```

```
curl --proxy http://127.0.0.1:7890 https://www.google.com   # 200
curl --proxy http://127.0.0.1:7890 https://github.com       # 200
curl --proxy http://127.0.0.1:7890 https://ifconfig.me      # <proxy exit IP, not local>
```

### DIFF
--- a/clash-lib/Cargo.toml
+++ b/clash-lib/Cargo.toml
@@ -68,8 +68,8 @@ rustls = { version = "0.23", default-features = false }
 webpki-roots = "1.0"
 
 # shadow-tls
-tokio-watfaq-rustls = { git = "https://github.com/Watfaq/tokio-rustls.git", rev = "638db32084d7ecf9c2660847b55d48d1186b4055", default-features = false, features = ["logging", "tls12"] }
-watfaq-rustls = { git = "https://github.com/Watfaq/rustls.git", rev = "4cae3aa2e84ea29d8a74b495793773bdb0a72206", default-features = false }
+tokio-watfaq-rustls = { git = "https://github.com/Watfaq/tokio-rustls.git", rev = "cf8961ac1a36e580d0e38bedc8a41ca4a9b301e8", default-features = false, features = ["logging", "tls12"] }
+watfaq-rustls = { git = "https://github.com/Watfaq/rustls.git", rev = "c3ab043d673029d245fd618b9bc86fd6a6109bae", default-features = false }
 
 # Error handing & logging
 thiserror = "2"
@@ -98,6 +98,7 @@ lru_time_cache = "0.11"
 uuid = { version = "1", features = ["v4", "fast-rng", "macro-diagnostics", "serde"] }
 network-interface = { version = "2" }
 base64 = "0.22"
+hex = "0.4"
 zstd = "0.13"
 portable-atomic = { version = "1", features = ["serde"] }
 
--- a/clash-lib/src/proxy/converters/vless.rs
+++ b/clash-lib/src/proxy/converters/vless.rs
@@ -3,7 +3,9 @@ use crate::{
     config::internal::proxy::OutboundVless,
     proxy::{
         HandlerCommonOptions,
-        transport::{GrpcClient, H2Client, TlsClient, WsClient},
+        transport::{
+            GrpcClient, H2Client, RealityClient, TlsClient, Transport, WsClient,
+        },
         vless::{Handler, HandlerOptions},
     },
 };
@@ -28,6 +30,69 @@ impl TryFrom<&OutboundVless> for Handler {
                 s.common_opts.server
             );
         }
+
+        if s.client_fingerprint.is_some() {
+            warn!(
+                "client-fingerprint (uTLS) is not yet implemented, ignored for {}",
+                s.common_opts.name
+            );
+        }
+
+        let tls: Option<Box<dyn Transport>> = if let Some(ref reality_opts) =
+            s.reality_opts
+        {
+            // vless with reality
+
+            // reality public-key bytes
+            let pk_bytes =
+                super::utils::decode_base64_public_key(&reality_opts.public_key)?;
+
+            // reality short id bytes
+            let short_id = super::utils::decode_short_id(&reality_opts.short_id)?;
+
+            // SNI
+            let sni = s
+                .server_name
+                .clone()
+                .unwrap_or_else(|| s.common_opts.server.clone());
+
+            Some(Box::new(RealityClient::new(sni, pk_bytes, short_id)))
+        } else {
+            // vless without reality
+            match s.tls.unwrap_or_default() {
+                true => {
+                    let client = TlsClient::new(
+                        s.skip_cert_verify.unwrap_or_default(),
+                        s.server_name.as_ref().map(|x| x.to_owned()).unwrap_or(
+                            s.ws_opts
+                                .as_ref()
+                                .and_then(|x| {
+                                    x.headers.clone().and_then(|x| {
+                                        let h = x.get("Host");
+                                        h.cloned()
+                                    })
+                                })
+                                .unwrap_or(s.common_opts.server.to_owned()),
+                        ),
+                        s.network
+                            .as_ref()
+                            .map(|x| match x.as_str() {
+                                "tcp" => Ok(vec![]),
+                                "ws" => Ok(vec!["http/1.1".to_owned()]),
+                                "http" => Ok(vec![]),
+                                "h2" | "grpc" => Ok(vec!["h2".to_owned()]),
+                                _ => Err(Error::InvalidConfig(format!(
+                                    "unsupported network: {x}"
+                                ))),
+                            })
+                            .transpose()?,
+                        None,
+                    );
+                    Some(Box::new(client))
+                }
+                false => None,
+            }
+        };
 
         Ok(Handler::new(HandlerOptions {
             name: s.common_opts.name.to_owned(),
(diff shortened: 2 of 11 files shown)


clash-lib/Cargo.toml
```diff
@@ -98,6 +98,7 @@ lru_time_cache = "0.11"
 uuid = { version = "1", features = ["v4", "fast-rng", "macro-diagnostics", "serde"] }
 network-interface = { version = "2" }
 base64 = "0.22"
+hex = "0.4"
```
ibigbug, line 101: fn from_hex<T: AsRef<[u8]>>(hex: T) -> Result<Self, Self::Error> {
        let hex = hex.as_ref();
        if hex.len() % 2 != 0 {
            return Err(FromHexError::OddLength);
        }

        hex.chunks(2)
            .enumerate()
            .map(|(i, pair)| Ok(val(pair[0], 2 * i)? << 4 | val(pair[1], 2 * i + 1)?))
            .collect()
    }


It seems we can quickly create a helper function for the decode

clash-lib/src/proxy/converters/vless.rs
```diff
@@ -29,6 +33,79 @@ impl TryFrom<&OutboundVless> for Handler {
             );
         }
 
+        if s.client_fingerprint.is_some() {
+            warn!(
+                "client-fingerprint (uTLS) is not yet implemented, ignored for {}",
```
ibigbug, line 36: Does the forked rustls still not support this?

SKTT1Ryze: Not yet, i think.

clash-lib/Cargo.toml
SKTT1Ryze: It's better to use third-party crates for utility functions rather than implementing them ourselves?

ibigbug: It's ok for this one I guess. Just had bad memories about the left pad incident so myself personally prefer not to add too many 3rd party deps.

clash-lib/src/proxy/vless/stream.rs
```diff
@@ -45,19 +47,27 @@ impl VlessStream {
...
         buf.put_slice(self.uuid.as_bytes());
-        buf.put_u8(0); // Additional info length (0 for simplicity)
+
+        if let Some(ref flow) = self.flow {
+            let addon = build_addon_bytes(flow);
+            buf.put_u8(addon.len() as u8);
+            buf.extend_from_slice(&addon);
+        } else {
+            buf.put_u8(0); // No addon
+        }
```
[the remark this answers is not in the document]
SKTT1Ryze: Checking length every time is not zero-cost

clash-lib/Cargo.toml
SKTT1Ryze: Fair point about avoiding unnecessary deps.But hex decoding is one of those cases where a stable, popular crate is more reliable than a quick custom implementation. :)

clash-lib/src/proxy/vless/vision.rs
```diff
@@ -0,0 +1,322 @@
...
+            framed.put_u16(padding_len);
+            for _ in 0..padding_len {
+                framed.put_u8(rand::random::<u8>());
+            }
+        }
+
+        framed.put_u8(VISION_TYPE_DATA);
+        framed.put_u16(data.len() as u16);
+        framed.extend_from_slice(data);
+
```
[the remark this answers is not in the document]
SKTT1Ryze: Checking length is not zero-cost, consider fixing this after some issues occured.

clash-lib/src/proxy/vless/stream.rs
```diff
@@ -203,3 +213,93 @@ impl AsyncWrite for VlessStream {
         Pin::new(&mut self.inner).poll_shutdown(cx)
     }
 }
+
+/// Encode the flow field as a Protobuf field-1 length-delimited value.
+/// Format: [0x0A][varint len][bytes]
+pub(crate) fn build_addon_bytes(flow: &str) -> Vec<u8> {
+    let mut buf = Vec::with_capacity(2 + flow.len());
+    buf.push(0x0A); // field 1, wire type 2 (length-delimited)
+    buf.push(flow.len() as u8); // single-byte varint (flow strings are short)
```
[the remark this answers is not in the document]
SKTT1Ryze: "xtls-rprx-vision" is the only possible value on `flow`.

clash-lib/src/proxy/vless/vision.rs
```diff
@@ -0,0 +1,322 @@
...
+impl AsyncWrite for VisionStream {
+    fn poll_write(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &[u8],
+    ) -> Poll<io::Result<usize>> {
+        let fut = self.write_framed(buf);
+        tokio::pin!(fut);
+        fut.poll(cx)
+    }
```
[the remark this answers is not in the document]
SKTT1Ryze: fixed.

clash-lib/src/proxy/vless/vision.rs
```diff
@@ -0,0 +1,322 @@
...
+            }
+
+            // Need more raw bytes from the inner stream.
+            let mut tmp = [0u8; 4096];
+            let mut tmp_buf = ReadBuf::new(&mut tmp);
+            match Pin::new(&mut self.inner).poll_read(cx, &mut tmp_buf) {
+                Poll::Ready(Ok(())) => {
+                    let filled = tmp_buf.filled();
+                    if filled.is_empty() {
+                        return Poll::Ready(Ok(())); // EOF
```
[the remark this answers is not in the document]
SKTT1Ryze: fixed.

ibigbug:
thanks!

PR #1070, opened 2026-03-21
fix(netstack): add SYN flood protection & deduplication

Alma-human:
## Fixes: Memory Leak & SYN Flood Vulnerability in netstack

When evaluating `clash-netstack` TCP listener, a major memory leak and DoS vulnerability was found during SYN packet handling.

### Root Cause
In `poll_packets`, a new `tcp::Socket` and `TcpStreamHandle` (with two 256KB ring buffers) are eagerly allocated **for every `packet.syn() && !packet.ack()`**. 
This has two critical consequences:
1. **Memory leak on retransmissions**: If a legitimate TCP SYN is retransmitted (due to packet loss or delay), a duplicate socket and a new 512KB buffer set are allocated for the exact same connection tuple, leaking memory continuously.
2. **SYN Flood DoS**: A rogue local application (or any attacker capable of injecting into the TUN interface) can send numerous spoofed SYN packets, exhausting gigabytes of memory instantly (10,000 SYNs = ~5GB RAM) causing immediate Out-of-Memory (OOM) crashes.

### Fix
Added a `HashMap` based connection tracker (`syn_tracker`) to `poll_packets`:
- **Deduplication**: Drops processing of duplicate SYN packets (within 60 seconds) for the same `(src_addr, dst_addr)` tuple, allowing the initial socket/smoltcp instance to naturally handle retransmissions.
- **Flood Protection**: Enforces a strict limit (10,000) on the number of concurrent tracked half-open SYN packets. If the threshold is reached, new incoming SYNs are silently dropped without allocating the expensive `Socket` and ring buffers, protecting the proxy layer from crashing.

This keeps memory stable and ensures standard proxy stability during high concurrency or malicious traffic bursts.

### DIFF
--- a/clash-netstack/src/tcp_listener.rs
+++ b/clash-netstack/src/tcp_listener.rs
@@ -232,6 +232,7 @@ impl TcpListener {
                         continue;
                     }
 
+                    // TODO: get rid of this stupid log
                     if syn_tracker.len() >= SYN_TRACK_MAX {
                         syn_drop_count += 1;
                         if syn_drop_count == 1


ibigbug:
@Alma-human address the PR comments

Alma-human:
我已经把 review comments 里的建议都改好并且 push 上去啦，感谢反馈，麻烦再帮我瞅瞅！😎🚀

Alma-human:
我已经把 review comments 里的建议都改好并且 push 上去啦，包括 TTL 调到了 60s，以及增加了 memory budget 的限制，还有 KB/KiB 的注释也修正了。感谢反馈，麻烦再帮我瞅瞅！😎🚀

clash-netstack/src/tcp_listener.rs
```diff
@@ -190,6 +215,38 @@ impl TcpListener {
...
+                        device_injector.send(frame).map_err(|e| {
+                            error!("Failed to inject retransmitted SYN packet: {e}");
+                            std::io::Error::other(
+                                "Failed to inject retransmitted SYN packet",
+                            )
+                        })?;
+                        continue;
+                    }
+
+                    if syn_tracker.len() >= SYN_TRACK_MAX {
```
ibigbug, line 236:
```suggestion
                    // TODO: get rid of this stupid log
                    if syn_tracker.len() >= SYN_TRACK_MAX {
```

PR #1309, opened 2026-04-12
fix(ss-inbound): loop on UDP decrypt error to re-register waker

ibigbug:
## Problem

When `poll_recv_from_with_ctrl` returns `Poll::Ready(Err(e))` — e.g. an SS2022 authentication failure caused by scan traffic or a brief password mismatch during a config refresh — the I/O waker registered by the preceding `poll_recv_from` call has already been consumed. The previous code returned `Poll::Pending` at that point without registering a new waker, **permanently suspending the UDP dispatch task (t1)**.

Consequence: no further UDP packets are ever read from the socket, `recv_q` fills to the OS buffer limit (~177 KB observed in production), and every subsequent UDP probe/relay times out.

The failure is silent after the first error — no crash, no log spam — making it hard to diagnose. The triggering event can be as short-lived as a single scan packet or the one-packet race window that occurs when clash-rs reloads its user list.

## Root cause (detailed)

```
// shadowsocks-rust ProxySocket::poll_recv_from_with_ctrl:
match self.decrypt_recv_buffer(...) {
    Ok(x)  => Poll::Ready(Ok(...)),
    Err(e) => Poll::Ready(Err(...)),   // <-- data read, waker gone
}
```

```
// clash-rs InboundShadowsocksDatagram::poll_next (before this fix):
match rv {
    Ok(...)  => Poll::Ready(Some(...)),
    Err(e)   => {
        Poll::Pending  // BUG: no waker registered → task sleeps forever
    }
}
```

## Fix

Wrap the receive/match in a loop. On a decrypt error we fall through to the next iteration:
- If more data is already queued in `recv_q`, the next `poll_recv_from_with_ctrl` call processes it immediately.

No behaviour change on the happy path.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2137,7 +2137,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "caef6056a5788d05d173cdc3c562ac28ae093828f851f69378b74e4e3d578e41"
 dependencies = [
  "heck",
- "indexmap 2.13.1",
+ "indexmap 1.9.3",
  "itertools 0.14.0",
  "proc-macro-crate",
  "proc-macro2",
@@ -2293,7 +2293,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


ibigbug:
@copilot review the changes and be extremely cautious about the added loop and try to under stand it if it's correct. you can check shadowssocks-rust source code if you need

ibigbug:
@copilot now review it thoroughly and fix cargo fmt too

PR #1313, opened 2026-04-13
fix(ss-inbound): propagate user key and client_session_id for multi-user SS2022 UDP responses

ibigbug:
## Problem

In multi-user `2022-blake3-aes-*` mode, the server's UDP response was always encrypted using the server's **iPSK** (identity PSK), but the SS2022 spec and sing-shadowsocks v0.2.x both expect the response to be encrypted with the **uPSK** (user PSK, i.e. the last key in the PSK list).

The result: any client using sing-shadowsocks (or another compliant multi-user SS2022 implementation) would see every UDP response fail with:
```
decrypt packet: cipher: message authentication failed
```

## Root cause

`ProxySocket::poll_send_to_with_ctrl` selects the user key correctly **when `control.user` is set**:

```rust
if let Some(ref user) = control.user {
    key = user.key();  // uPSK
}
encrypt_server_payload(..., key, ...)
```

But `InboundShadowsocksDatagram` initialised `control` once at construction and never updated it from incoming packets, so `control.user` was always `None` and every response used the server iPSK instead.

The same omission left `control.client_session_id` at a stale random value, so the echoed client session ID in the response never matched the sender's actual session ID.

## Fix

In `poll_next`, after successfully receiving a packet, copy the sender's `client_session_id` and `user` into `self.control` before returning the `UdpPacket` to the dispatcher. When `poll_flush` sends the response, `self.control` now carries the correct user context.

Also initialise `server_session_id` to a random value at construction instead of leaving it at zero.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10545,7 +10545,7 @@ dependencies = [
 
 [[package]]
 name = "watfaq-netstack"
-version = "0.1.1"
+version = "0.1.0"
 dependencies = [
  "bytes",
  "env_logger",
--- a/clash-lib/src/proxy/shadowsocks/inbound/datagram.rs
+++ b/clash-lib/src/proxy/shadowsocks/inbound/datagram.rs
@@ -2,7 +2,7 @@ use crate::{
     common::errors::new_io_error, proxy::datagram::UdpPacket, session::SocksAddr,
 };
 use futures::ready;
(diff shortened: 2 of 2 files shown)


ibigbug:
@copilot review the PR and check if there could be any concurrency issues

PR #1319, opened 2026-04-19
fix: compat mode

### LINKED ISSUE #1318: Compatibility mode enabled by default forcibly overrides user-configured `mmdb` value
In version v0.9.7, the `--compatibility` parameter defaults to `true`. When this mode is enabled, it **unconditionally overrides** the user's configured `mmdb` and `geosite` values, causing custom mmdb paths to fail.

## Steps to Reproduce

1. Create a configuration file `config.yaml`:

   ```yaml
   mmdb: GeoLite2-Country.mmdb
   mmdb-download-url: https://fastly.jsdelivr.net/gh/Loyalsoldier/geoip@release/GeoLite2-Country.mmdb
   ```

2. Ensure the local file `GeoLite2-Country.mmdb` exists

3. Start clash-rs without any additional parameters:

   ```bash
   ./clash-rs -c config.yaml
   ```

## Actual Behavior

The program ignores the user's `mmdb: GeoLite2-Country.mmdb` configuration and forcibly uses `Country.mmdb`, resulting in:

1. Local `Country.mmdb` file not being found
2. Triggering a download from `mmdb-download-url`
3. The downloaded file is saved as `Country.mmdb` instead of the user's expected `GeoLite2-Country.mmdb`

## Root Cause

In `clash-bin/src/main.rs`, line 189:

```rust
if cli.compatibility {
    // ...
    config.general.mmdb = Some("Country.mmdb".to_string());  // Unconditional override!
    config.general.geosite = Some("geosite.dat".to_string());
}
```

At the same time, the default value of `--compatibility` is `true` (line 82), meaning even if the user does not explicitly enable compatibility mode, the override logic is still triggered.

### Option 1: Set default value only when not configured (Recommended)

```rust
if cli.compatibility {
    // Only set defaults if the user has not configured them
    if config.general.mmdb.is_none() {
        config.general.mmdb = Some("Country.mmdb".to_string());
    }
    if config.general.geosite.is_none() {
        config.general.geosite = Some("geosite.dat".to_string());
    }
}
```

### Option 2: Change compatibility default to false

If compatibility mode does indeed change behavior and may cause issues, it is recommended to change the default to `false` and let users actively choose to enable it.

### Option 3: Add an independent configuration item

Allow setting `compatibility-mode: true/false` in the configuration file, which takes priority over the command-line parameter.

## Environment Information

- **Version:** v0.9.7
- **OS:** macOS
- **Usage Scenario:** Launched via direct command line

Itsusinn:
This pull request introduces a conditional block to improve compatibility mode handling in the application's startup logic. Now, when compatibility mode is enabled, the code ensures that default values are set for certain configuration fields if they are not already specified.

Configuration compatibility enhancements:

* In `clash-bin/src/main.rs`, when the `compatibility` flag is set, the code checks if `config.general.mmdb` and `config.general.geosite` are unset, and assigns them default values (`"Country.mmdb"` and `"geosite.dat"`, respectively) if needed.

close #1318

## Summary by CodeRabbit

* **Bug Fixes**
  * Compatibility mode no longer overwrites existing configuration values with defaults; default files are only applied when those settings are empty, preserving any user-provided paths.

### DIFF
--- a/clash-bin/src/main.rs
+++ b/clash-bin/src/main.rs
@@ -186,13 +186,11 @@ fn main() -> anyhow::Result<()> {
         if let Some(dir) = &cli.directory {
             std::env::set_current_dir(dir)?;
         }
-        if cli.compatibility {
-            if config.general.mmdb.is_none() {
-                config.general.mmdb = Some("Country.mmdb".to_string());
-            }
-            if config.general.geosite.is_none() {
-                config.general.geosite = Some("geosite.dat".to_string());
-            }
+        if config.general.mmdb.is_none() {
+            config.general.mmdb = Some("Country.mmdb".to_string());
+        }
+        if config.general.geosite.is_none() {
+            config.general.geosite = Some("geosite.dat".to_string());
         }
     }
 
--- a/clash-bin/src/main.rs
+++ b/clash-bin/src/main.rs
@@ -186,8 +186,14 @@ fn main() -> anyhow::Result<()> {
         if let Some(dir) = &cli.directory {
             std::env::set_current_dir(dir)?;
         }
-        config.general.mmdb = Some("Country.mmdb".to_string());
-        config.general.geosite = Some("geosite.dat".to_string());
+        if cli.compatibility {
+            if config.general.mmdb.is_none() {
+                config.general.mmdb = Some("Country.mmdb".to_string());
+            }
+            if config.general.geosite.is_none() {
+                config.general.geosite = Some("geosite.dat".to_string());
+            }
+        }
     }
 
     clash::start_scaffold(clash::Options {


clash-bin/src/main.rs
```diff
@@ -186,8 +186,14 @@ fn main() -> anyhow::Result<()> {
         if let Some(dir) = &cli.directory {
             std::env::set_current_dir(dir)?;
         }
-        config.general.mmdb = Some("Country.mmdb".to_string());
-        config.general.geosite = Some("geosite.dat".to_string());
+        if cli.compatibility {
```
ibigbug: Duplicated check ?

Itsusinn: Yeah, somehow i wrote this, due to branch conflict probably

ibigbug: can you fix this?

Itsusinn: working...could you fix it for me?

ibigbug: @copilot fix the nested double check here and make sure cargo +nightly fmt and clippy clean

ibigbug (approved):

PR #1330, opened 2026-04-24
fix: ss2022 throughput, simple-obfs TLS state machine, MMDB path

ibigbug:
### What does this PR do?

Fixes several bugs causing SS2022 outbound to be slow or fail (decrypt errors, low throughput) while DIRECT was full speed.

#### Bug fixes

**simple-obfs TLS: data loss on `Poll::Pending` (root cause of decrypt failures)**

The original `reading()` helper created `read_exact` futures with stack-local buffers. When `read_exact` returned `Poll::Pending` after consuming M bytes from the inner stream, the future and its buffer were dropped. The next `poll_read` restarted from byte 0 — M bytes were permanently lost. This desynchronised the TLS frame parser, causing AEAD tag mismatches and wildly fluctuating speeds.

Fix: rewrote `poll_read` / `poll_write` as explicit state machines (`ReadState` enum, `write_buf`/`write_pos`/`write_committed` fields) whose buffers live in the struct and survive `Poll::Pending`. Also replaced `Vec<u8>` scratch buffers in `SkippingHeader` with a fixed-size `[u8; 105]` array to avoid per-record heap allocations.

**SS2022: undersized copy buffer**

`DEFAULT_BUFFER_SIZE` was 16 KB — 4× smaller than SS2022's `MAX_PACKET_SIZE` of 65535 bytes. Increased to 64 KB.

**SS2022 UDP: per-packet `Context` allocation and recv buffer**

`Context::new_shared()` was called on every UDP send. Changed to a shared `Arc<Context>` per `Handler`. Also pre-allocated a 65535-byte `recv_buf` in the UDP datagram wrapper instead of allocating per packet.

**MMDB: path resolution after `set_current_dir`**

`set_current_dir(relative_path)` changed the cwd, then the same relative path was used again, doubling it. Fix: canonicalize the path before calling `set_current_dir`.

### Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring / cleanup
- [ ] Other

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed compatibility-mode directory path handling
  * Improved TLS obfuscation state persistence and reliability

* **Performance Improvements**
  * Increased UDP buffering to reduce drops under load
  * Reduced per-packet UDP allocation overhead

* **Refactor**
  * Reused connection context and simplified synchronization for outbound proxying

### DIFF
--- a/clash-bin/src/main.rs
+++ b/clash-bin/src/main.rs
@@ -184,7 +184,15 @@ fn main() -> anyhow::Result<()> {
              recommended to enable this if you are using clash verge."
         );
         if let Some(dir) = &cli.directory {
-            std::env::set_current_dir(dir)?;
+            // Canonicalize to an absolute path before changing the process cwd.
+            // If `dir` is relative (e.g. `./clash-bin/tests/data/config`),
+            // calling set_current_dir then passing the same relative string as
+            // `cwd` to start_scaffold would cause paths like
+            // `cwd.join("Country.mmdb")` to be resolved from the *new* process
+            // cwd, doubling the directory segments and producing a path that
+            // doesn't exist (os error 2).
+            let abs = std::fs::canonicalize(dir)?;
+            std::env::set_current_dir(&abs)?;
         }
         if config.general.mmdb.is_none() {
             config.general.mmdb = Some("Country.mmdb".to_string());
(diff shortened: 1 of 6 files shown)


ibigbug:
@copilot update PR title and description accordingly

ibigbug:
@copilot fix the PR title and description again it was overriden by some one. also mention the cli compatibility fix

PR #1331, opened 2026-04-25
feat(cli): align config validation with mihomo behavior

Tunglies:
- **feat(cli): align config test with mihomo**
- **fix(config): align provider health-check defaults**

对其 mihomo 对 -t 测试行为输出格式
对其 mihomo 对字段 `health-check` 填充和检测

## Summary by CodeRabbit

* **New Features**
  * Support for base64-encoded config input via CLI and env vars; config can come from multiple sources and is resolved robustly.
  * Config loading now creates a default config file only when a missing file is chosen as the source.

* **Bug Fixes**
  * Improved error handling and centralized logging for config resolution/parsing.
  * Proxy provider health-check behavior now uses effective/defaulted interval and laziness values.

* **Tests**
  * Added integration tests covering file, stdin, and base64 config inputs, error cases, and default-file creation.

* **Chores**
  * Added new runtime and dev dependencies.


clash-lib/src/config/internal/proxy.rs
```diff
@@ -697,17 +698,52 @@ pub struct OutboundFileProvider {
     pub name: String,
     pub path: String,
     pub interval: Option<u64>,
+    #[serde(default = "default_provider_health_check")]
```
ibigbug: would impl Default for HealthCheck simplify it here a bit?

clash-lib/src/config/internal/proxy.rs
```diff
@@ -697,17 +698,52 @@ pub struct OutboundFileProvider {
...
-#[derive(serde::Serialize, serde::Deserialize, Debug)]
+#[derive(serde::Serialize, serde::Deserialize, Debug, Clone)]
 pub struct HealthCheck {
+    #[serde(default)]
     pub enable: bool,
+    #[serde(default)]
     pub url: String,
+    #[serde(default)]
     pub interval: u64,
+    #[serde(default = "default_provider_health_check_lazy")]
```
ibigbug: i think there's a `default_bool_true` in clash-lib can you use that instead so. you don't need to do Option here

clash-bin/src/main.rs
```diff
@@ -111,43 +114,28 @@ fn main() -> anyhow::Result<()> {
...
-                exit(1);
-            }
-        };
-
-        if config_file.write_all(default_config.as_bytes()).is_err() {
-            eprintln!("default profile cannot be written: {file}");
+    let config_input = match ConfigInput::resolve(&cli) {
+        Ok(config_input) => config_input,
+        Err(err) => {
+            print_mihomo_log("fatal", &err.to_string());
```
ibigbug: it looks like print_mihomo_log is default behavior now with no other option, maybe you can remove the "mihomo" from the function names everywhere, if that simplifies code and logic in any way.

or otherwise you can check the cli.compatiblity flag to enable regarding mihomo behaviors but i think that might be more involved

clash-bin/src/main.rs
```diff
@@ -221,3 +209,209 @@ fn main() -> anyhow::Result<()> {
...
+                clash::Config::File(path.to_string_lossy().to_string()).try_parse()
+            }
+            Self::Bytes { content } => {
+                clash::Config::Str(content.clone()).try_parse()
+            }
+        }
+    }
+}
+
+fn preprocess_args(args: impl IntoIterator<Item = String>) -> Vec<String> {
```
ibigbug, line 302: can you add a test for this so that 

```
$ clash-rs -d DIR -f test.yml
$ clash-rs -d DIR # no -f so just find config.yaml under DIR
$ clash-rs -f test.yaml # find test.yaml in current dir
$ clash-rs # find config.yaml under current dir
```

all work?

Tunglies:
It works
```bash
> pwd
/Users/user/workspace/clash-rs

> cargo run -p clash-rs -- -d /Users/user/Library/Application\ Support/io.github.clash-verge-rev.clash-verge-rev -f clash-verge.yaml
    Finished [`dev` profile [unoptimized + debuginfo]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 0.44s
     Running `target/debug/clash-rs -d '/Users/tunglies/Library/Application Support/io.github.clash-verge-rev.clash-verge-rev' -f clash-verge.yaml`
Compatibility mode enabled. This may cause some issues, but it is recommended to enable this if you are using clash verge.
26-04-25 15:11:51:132242 DEBUG ThreadId(01) clash_lib: clash-lib/src/lib.rs:366: tun enabled, initializing default outbound interface
26-04-25 15:11:51:139308 DEBUG ThreadId(01) clash_lib: clash-lib/src/lib.rs:372: initializing cache store
```

ibigbug (approved):

ibigbug:
please fix the failed tests

ibigbug (commented):
## Code review

### �� High: Health check with empty URL will spam errors

When `health-check: {enable: true}` is set without a `url`, `effective_interval()` returns 300 (the default) and the health check kicks off every 5 minutes — but fires against an empty string URL, which fails to parse on every tick.

```rust
pub fn effective_interval(&self) -> u64 {
    if !self.enable {
        return 0;
    }
    if self.interval == 0 {
        300  // <-- fires even when url is ""
    } else {
        self.interval
    }
}
```

**Suggested fix** — treat a missing URL the same as disabled:

```rust
pub fn effective_interval(&self) -> u64 {
    if !self.enable || self.url.is_empty() {
        return 0;
    }
    if self.interval == 0 { 300 } else { self.interval }
}
```

---

### ⚠️ Medium: `-config` flag semantics differ from mihomo

In mihomo, `-config` specifies a **config file path**. Here it is remapped to `--config-string` (base64 input):

```rust
"-config" => "--config-string".to_string(),
```

Anyone currently using `-config /path/to/config.yaml` would silently get a base64-decode error instead. If the goal is mihomo CLI compatibility, `-config` should map to `--config` (file path), not `--config-string`.

---

The test failures in CI are addressed by PR #1348 (skip tests in cross/QEMU envs where subprocess exec returns 127, fix Windows backslash escaping in `empty_file_reports_empty_file_error`).

ibigbug:
@Tunglies check the AI review comment?

PR #1333, opened 2026-04-26
fix: support domain in proxy-server-nameserver

Itsusinn:
This pull request improves the handling and validation of the `proxy-server-nameserver` configuration in the DNS resolver. The changes ensure that domain hostnames are properly supported, empty or unusable nameserver lists are detected early, and the system gracefully falls back or errors out when no valid DNS clients are available.

**Validation and error handling improvements:**

* Updated the parsing logic in `Config` to allow domain hostnames for `proxy-server-nameserver`, and now returns a clear error if no usable entries are found after parsing.
* In `EnhancedResolver`, added a check to log a warning and fall back to main nameservers if no proxy-server-nameserver clients are initialized, instead of proceeding with an empty list.
* Added a guard in the DNS query method to return an explicit error if there are no DNS clients available for a query, preventing potential panics or undefined behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * DNS nameserver validation now accepts domain hostnames in addition to IP addresses for proxy server configurations.
  * Improved error messaging when nameserver entries fail to parse.
  * Enhanced fallback behavior: if proxy-server nameserver initialization fails, DNS resolution automatically reverts to main nameservers for improved reliability.

### DIFF
--- a/clash-lib/src/app/dns/resolver/enhanced.rs
+++ b/clash-lib/src/app/dns/resolver/enhanced.rs
@@ -130,16 +130,24 @@ impl EnhancedResolver {
 
         let proxy_resolver =
             if let Some(proxy_resolver) = cfg.proxy_server_nameserver {
-                Some(
-                    make_clients(
-                        proxy_resolver,
-                        None,
-                        Arc::new(RwLock::new(std::collections::HashMap::new())),
-                        edns_client_subnet.clone(),
-                        cfg.fw_mark,
-                    )
-                    .await,
+                let clients = make_clients(
+                    proxy_resolver,
+                    Some(default_resolver.clone()),
+                    outbounds.clone(),
+                    edns_client_subnet.clone(),
+                    cfg.fw_mark,
                 )
+                .await;
+                if clients.is_empty() {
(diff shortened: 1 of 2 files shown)


clash-lib/src/app/dns/resolver/enhanced.rs
```diff
@@ -130,16 +130,24 @@ impl EnhancedResolver {
...
-                        None,
-                        Arc::new(RwLock::new(std::collections::HashMap::new())),
-                        edns_client_subnet.clone(),
-                        cfg.fw_mark,
-                    )
-                    .await,
+                let clients = make_clients(
+                    proxy_resolver,
+                    Some(default_resolver.clone()),
+                    outbounds.clone(),
```
ibigbug, line 136: What would happen if you use proxy to resolve proxy servers? Infinite recursion?

Itsusinn: Yes, But no guarantee to no recursion is OKAY to me.

ibigbug (approved):

PR #1347, opened 2026-04-27
fix: verge compatibility part 4

Itsusinn:


clash-lib/src/app/api/handlers/provider.rs
```diff
@@ -33,7 +33,7 @@ struct ProviderState {
 pub fn routes(outbound_manager: ThreadSafeOutboundManager) -> Router<Arc<AppState>> {
     let state = ProviderState { outbound_manager };
     Router::new()
-        .route("/", get(get_providers))
+        .route("/", any(get_providers))
```
ibigbug, line 36: why is this any?

Itsusinn: Seems mihomo doesn't restrcit HTTP method. Could also be my misunderstand of tauri ipc mechanism

PR #1350, opened 2026-04-28
refactor(clash-lib): reduce nesting with if-let chains and guard clauses

Itsusinn:
- app/router/mod.rs: extract ASN lookup into dedicated method with guard clause; eliminate nested if-let/match pattern in match_route
- app/dns/resolver/enhanced.rs: flatten exchange() with early return guard and if-let chain; simplify lookup_ip() with propagation
- app/outbound/manager.rs: deduplicate 6 nearly identical match arms in load_group_outbounds via build_group_providers helper; extract load_proxy_providers boilerplate into make_proxy_set_provider; flatten initialization loop with if-let replace match
- cargo +nightly fmt applied

Assisted-by: OpenClaw:deepseek/deepseek-v4-flash

### Type

- [ ] Bug fix
- [ ] New feature
- [x] Refactoring / cleanup
- [ ] Other

### Checklist

## Summary by CodeRabbit

* **Refactor**
  * Simplified DNS lookup flow and cache-hit handling for a clearer, more direct resolution path with consistent cache fallback.
  * Consolidated proxy/provider construction and outbound group initialization to remove duplication, simplify error handling, and improve logging for failed providers.
  * Extracted ASN/country enrichment into a dedicated helper to streamline route matching and early-exit behavior when data is missing.


clash-lib/src/app/router/mod.rs
```diff
@@ -138,6 +124,35 @@ impl Router {
         (MATCH, None)
     }
 
+    /// Look up ASN for an IP address. Uses the simplified mmdb lookup first,
+    /// falling back to the full ASN lookup if the simplified one fails.
+    fn lookup_asn_for_ip(
```
ibigbug:
```suggestion
    fn populate_asn_for_ip(
```

ibigbug (approved):
lgtm bar the function name and clippy annotation

PR #1385, opened 2026-05-09
fix: use child_token() instead of clone() for reload API server cancellation token

xiguagua:
CancellationToken::clone() shares cancellation state, so shutting down the old API server on reload would also cancel the new one's token, causing delay/healthcheck endpoints to become unreachable after the second hot-reload cycle.

### Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring / cleanup
- [ ] Other

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling during configuration reloads to reduce resource leaks and increase stability of the API server lifecycle.

* **Tests**
  * Updated unit tests to use local mock endpoints and more reliable timing checks, improving test robustness.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Watfaq/clash-rs/pull/1385)

### DIFF
--- a/clash-lib/src/app/remote_content_manager/mod.rs
+++ b/clash-lib/src/app/remote_content_manager/mod.rs
@@ -1001,15 +1001,24 @@ mod tests {
         tests::initialize,
     };
     use futures::TryFutureExt;
+    use httpmock::{Method::GET, MockServer};
     use std::{net::Ipv4Addr, sync::Arc, time::Duration};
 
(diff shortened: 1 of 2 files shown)


ibigbug:
thank you - i think the fix is correct but tests are genuinely broken, maybe relied on the previous behaviour. could you plz help to take a look too?

ibigbug (approved):

PR #1393, opened 2026-05-10
fix(direct): align UDP bind_addr family with family_hint destination

ibigbug:
## Root cause

After the IPv4-mapped canonicalization fix (commit 6783909), `sess.source` for a client connecting as `::ffff:x.x.x.x` is now the canonical `x.x.x.x`. So `sess.source.is_ipv4()` returns `true` and `bind_addr` is set to `0.0.0.0`.

However, `family_hint_for_session` is based on the **destination**, not the source. When the destination is IPv6, `new_udp_socket` creates an `AF_INET6` socket — then tries to `bind(0.0.0.0:0)` → **`EAFNOSUPPORT` (os error 97)**.

This is why IPv6 UDP probes fail when clients are on IPv4: the canonicalized source makes the DIRECT handler pick an IPv4 bind address, but the destination demands an IPv6 socket.

## Fix

Derive `bind_addr` from `family_hint` first (destination-driven), and only fall back to `sess.source` when there is no hint. This ensures bind address and socket domain always agree.

## Related

- #1392 — fixes the control-entry key mismatch caused by the same canonicalization (IPv4 UDP drops)
- This PR fixes the subsequent IPv6 UDP EAFNOSUPPORT caused by the same root change

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UDP connection handling to correctly bind to the appropriate IP address family (IPv4 or IPv6), enhancing connectivity reliability in mixed network environments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Watfaq/clash-rs/pull/1393)

### DIFF
--- a/clash-lib/src/proxy/direct/mod.rs
+++ b/clash-lib/src/proxy/direct/mod.rs
@@ -94,10 +94,14 @@ impl OutboundHandler for Handler {
         resolver: ThreadSafeDNSResolver,
     ) -> std::io::Result<BoxedChainedDatagram> {
         let family_hint = family_hint_for_session(sess, &resolver).await;
-        let bind_addr: std::net::IpAddr = if sess.source.is_ipv4() {
-            std::net::Ipv4Addr::UNSPECIFIED.into()
-        } else {
-            std::net::Ipv6Addr::UNSPECIFIED.into()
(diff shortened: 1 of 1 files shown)


ibigbug:
Closing — folding this fix into #1392 since both address the same root cause.

PR #1400, opened 2026-05-13
fix: gracefully handle stale interface binding with fallback to default route

xiguagua:
iOS returns EADDRNOTAVAIL (os error 49) when bind_device_by_index_v4/v6 is called with a stale interface index (after network changes like Wi-Fi switches or VPN connect/disconnect). Similarly, Linux bind_device returns ENODEV when the interface name no longer exists.

Rename must_bind_socket_on_interface -> bind_socket_on_interface to reflect that binding is now best-effort rather than mandatory. When the target interface is stale (AddrNotAvailable/NotFound), log a warning and fall back to the default route instead of failing the connection.

### Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring / cleanup
- [ ] Other

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network interface binding reliability on Apple and Unix platforms by gracefully handling stale interface indices.
  * Fixed log file creation to properly create missing files instead of failing.

* **New Features**
  * Build script now supports flexible iOS/macOS build targeting via a `--full` flag.

* **Chores**
  * Updated dependencies for enhanced security and observability.
  * Removed memory field from statistics output.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Watfaq/clash-rs/pull/1400)


ibigbug:
Wouldn't the connection fail anyways if it's supposed to go out via a specific interface when fallback to default?

xiguagua:
> Wouldn't the connection fail anyways if it's supposed to go out via a specific interface when fallback to default?

That's a different issue. This PR addresses an issue where interface binding fails after a device has been idle for a period of time on the iOS platform. causing endpoint speed tests to fail.

ibigbug:
In that case I think you should properly stop and resume on network path changed and populate the active interface again. 

This PR won't fix the issue it just hides it and resurface with a different error instead, right?

xiguagua:
> In that case I think you should properly stop and resume on network path changed and populate the active interface again.
> 
> This PR won't fix the issue it just hides it and resurface with a different error instead, right?

It is not hiding，just let system choose. 
resurface with a error not associated

ibigbug:
It's probably fine and works on iOS where NE protects the sockets. 

But it's logically wrong when you decided to bind an interface for a socket but it ended up going through another interface.

xiguagua:
Got it. 
Would make another PR

PR #1403, opened 2026-05-13
fix: gracefully handle stale interface binding on Apple platforms

xiguagua:
macOS returns EADDRNOTAVAIL (os error 49) when bind_device_by_index_v4/v6 is called with a stale interface index after network changes (Wi-Fi switch, VPN connect/disconnect, etc.). Extract handle_bind_result helper to DRY the AddrNotAvailable fallback logic across IPV4/IPV6 branches.

### Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring / cleanup
- [ ] Other

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network-interface binding on Apple platforms: when an interface index is stale, the app now logs a warning and gracefully falls back to the default route instead of failing, reducing connection errors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Watfaq/clash-rs/pull/1403)

### DIFF
--- a/clash-lib/src/proxy/utils/platform/apple.rs
+++ b/clash-lib/src/proxy/utils/platform/apple.rs
@@ -17,16 +17,30 @@ pub(crate) fn must_bind_socket_on_interface(
         );
         return Ok(());
     }
-    match family {
+    let result = match family {
         socket2::Domain::IPV4 => {
             socket.bind_device_by_index_v4(std::num::NonZeroU32::new(index))
         }
         socket2::Domain::IPV6 => {
             socket.bind_device_by_index_v6(std::num::NonZeroU32::new(index))
         }
-        _ => Err(io::Error::new(
-            io::ErrorKind::InvalidInput,
-            "unsupported socket family",
-        )),
-    }
+        _ => {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "unsupported socket family",
+            ));
+        }
+    };
+    result.or_else(|e| {
+        if e.kind() == io::ErrorKind::AddrNotAvailable {
+            warn!(
+                "stale interface index {index} for '{}', \
+                 falling back to default route: {e}",
+                iface.name
+            );
+            Ok(())
+        } else {
+            Err(e)
+        }
+    })
 }


clash-lib/src/proxy/utils/platform/apple.rs
```diff
@@ -18,15 +36,19 @@ pub(crate) fn must_bind_socket_on_interface(
         return Ok(());
     }
     match family {
-        socket2::Domain::IPV4 => {
-            socket.bind_device_by_index_v4(std::num::NonZeroU32::new(index))
-        }
-        socket2::Domain::IPV6 => {
-            socket.bind_device_by_index_v6(std::num::NonZeroU32::new(index))
-        }
+        socket2::Domain::IPV4 => handle_bind_result(
```
ibigbug: i think this can be simplified by 

```
let error = if socket.v4 {} else if socket.v6 {} else { return io error;}

if error.kind() == stale iface { ok } else { err }

ibigbug: but i genuinely think this is uncessary, as you should listen to the network path satisfication events and handle that properly, e.g. shutdown on offline and restart on device change etc.

xiguagua: Oh, we are talking about iOS, and we are doing cross-language invoke, NE process behavior is hard to reason about.

ibigbug: there is an iOS API in NE that you can use to listen to the network path change, I believe.

xiguagua: Yeah，clash ffi still give two apis (start and shutdown). 
What I could do？shutdown and start when there is change? 
That is not just down and up, because SOT is on app side, app is not alway on like NE process
I think that is worse.

ibigbug: why do you think it's worse - when network is offline, I think clash should be shutdown too, and it won't cost too much to start again given all assets are downloaded, maybe providers initialization could use a cache too, but it's no point running clash when there's no network right?

ideally clash can listen to the network interface change, but it's too platform specific so I haven't integrated that feature.

xiguagua: I thought clash as a long-run background service , and iOS give NE process alive for this reason.
And I see clash provide web api /reload for config-refresh, not asking invoker to shutdown and restart

xiguagua: Or maybe we could add an new http api to clashrs, update current active network interface

ibigbug: NE allows background VPN service doesn't necessarily mean you have to make it alive all the time, also for power consumption reason you should shutdown clash too when network is offline

xiguagua: 我英文不太好，词不达意，直接说中文了。

就如你所说，clashrs 不止于服务 apple 平台，我开发的 壳app 也不只是采用 clashrs 一个核心，不同核心的实现细节不同，clashrs 开关或许很快，但 mihomo 不是这样，不能为了 clashrs 这么做。我在上面已经提出一个方案，还请再次考虑

ibigbug: https://t.me/thisisnotclash can yiu join this

xiguagua: 偏题了。聚焦 must_bind_socket_on_interface 这个方法，目前发现设备闲置一段时候后，会调用失败，导致测速无效。按你提的建议监听系统事件 fix，那之后 must_bind_socket_on_interface 只要因其他问题返回error，网络连接还是会出问题，无法建立连接

PR #1404, opened 2026-05-13
fix: timeout on hyper handshake and consolidate proxy state writes

xmh0511:
Concurrent latency testing (via `/ui`) and proxy switching (via nyanpasu or any external client) could freeze all API responses and proxy traffic indefinitely, requiring a restart. No panic — pure liveness failure from compounding async lock misuse.

## Root cause chain

Tokio's `RwLock` is write-preferring: one pending writer blocks all subsequent readers. The hang followed this sequence:

1. `GET /providers/{name}/healthcheck` → `provider.healthcheck()` held the `ProxySetProvider.inner` **L2 read lock** across the entire 5–15s network probe phase (Rust temporary lifetime rule kept an unnamed `RwLockReadGuard` alive through a chained `.await`)
2. If any proxy accepted TCP but stalled before responding to HTTP, `hyper::handshake()` had **no timeout** → L2 read held **forever**
3. Fetcher's periodic pull attempted L2 **write** → blocked → L2 became write-pending
4. Every subsequent L2 read (routing decisions, `PUT /proxies/{group}`, all API handlers touching the provider) → **blocked**

## Fixes

**`proxy_set_provider.rs`** — drop lock before long awaits
- `healthcheck()` and `touch()`: clone `Arc<HealthCheck>` so the L2 read guard drops at the semicolon, then await with no lock held
  ```rust
  // Before — L2 read held for entire check duration
  self.inner.read().await.hc.check().await;
  // After — L2 read released immediately
  let hc = self.inner.read().await.hc.clone();
  hc.check().await;
  ```
- Updater closure: scope the L2 write guard to drop it before `hc.update(input).await`

**`mod.rs` — `url_test()`**
- Wrap both `hyper::client::conn::http1::handshake()` calls (HTTP + HTTPS) in `tokio::time::timeout(timeout, ...)` — without this, a stalling proxy turns the L2 hold into a permanent hang
- Combine the two sequential `proxy_state` write-lock acquisitions (`report_alive` + delay history append) into one, reducing write-burst pressure when many probes finish simultaneously

**`healthcheck.rs` — `kick_off()`**
- Re-read `inner.proxies` inside the tick loop instead of capturing a stale clone at spawn time; previously, provider updates (fetcher pulls) were invisible to the background health-check task

**`provider_helper.rs`**
- Separate the single L1 `ThreadSafeProxyProvider` read guard into two independent acquisitions so `touch()` and `proxies()` don't hold L1 across each other

**`provider.rs` API handlers**
- `provider_healthcheck`: spawn the check as a background task and return `202` immediately — handler no longer holds L1 for the probe duration
- `update_provider`: extract provider name from the same guard used for `update()`, removing an accidental double lock acquisition

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable proxy connection timeouts to avoid spurious hangs during URL tests.
  * Clearer, distinct error messages for handshake failures to aid troubleshooting.
  * Consolidated proxy liveness and delay recording for more consistent connectivity status and trimmed delay history to keep recent performance metrics accurate.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Watfaq/clash-rs/pull/1404)

### DIFF
--- a/clash-lib/src/app/remote_content_manager/mod.rs
+++ b/clash-lib/src/app/remote_content_manager/mod.rs
@@ -167,10 +167,21 @@ impl ProxyManager {
             .unwrap_or(true) // if not found, assume it's alive
     }
 
-    pub async fn report_alive(&self, name: &str, alive: bool) {
+    pub async fn report_alive(
+        &self,
+        name: &str,
+        alive: bool,
+        history: Option<DelayHistory>,
+    ) {
         let mut state = self.proxy_state.write().await;
-        let state = state.entry(name.to_owned()).or_default();
-        state.alive.store(alive, Ordering::Relaxed)
+        let entry = state.entry(name.to_owned()).or_default();
+        entry.alive.store(alive, Ordering::Relaxed);
+        if let Some(ins) = history {
+            entry.delay_history.push_back(ins);
+            if entry.delay_history.len() > 10 {
+                entry.delay_history.pop_front();
+            }
+        }
     }
 
     pub async fn delay_history(&self, name: &str) -> Vec<DelayHistory> {
@@ -747,10 +758,15 @@ impl ProxyManager {
             let resp = match uri.scheme() {
                 Some(scheme) if scheme == &http::uri::Scheme::HTTP => {
                     let io = TokioIo::new(stream);
-                    let (mut sender, conn) =
-                        hyper::client::conn::http1::handshake(io).await.map_err(
-                            |e| new_io_error(format!("failed to handshake: {e}")),
-                        )?;
+                    let (mut sender, conn) = tokio::time::timeout(
+                        timeout,
+                        hyper::client::conn::http1::handshake(io),
+                    )
+                    .await
+                    .map_err(|_| new_io_error("HTTP handshake timed out"))?
+                    .map_err(|e| {
+                        new_io_error(format!("failed to handshake: {e}"))
+                    })?;
 
                     tokio::task::spawn(async move {
                         if let Err(err) = conn.await {
@@ -786,10 +802,15 @@ impl ProxyManager {
 
                     let io = TokioIo::new(stream);
 
-                    let (mut sender, conn) =
-                        hyper::client::conn::http1::handshake(io).await.map_err(
-                            |e| new_io_error(format!("failed to handshake: {e}")),
-                        )?;
+                    let (mut sender, conn) = tokio::time::timeout(
+                        timeout,
+                        hyper::client::conn::http1::handshake(io),
+                    )
+                    .await
+                    .map_err(|_| new_io_error("HTTPS handshake timed out"))?
+                    .map_err(|e| {
+                        new_io_error(format!("failed to handshake: {e}"))
+                    })?;
 
                     tokio::task::spawn(async move {
                         if let Err(err) = conn.await {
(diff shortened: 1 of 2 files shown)


Itsusinn:
Even fix this deadlock, I think further LLM code could break it again

ibigbug:
i think the PR raised fair points that there could be a few dead lock issues, some of which are caused by design patterns. let me make a full revisit and do another fix

Itsusinn:
> i think the PR raised fair points that there could be a few dead lock issues, some of which are caused by design patterns. let me make a full revisit and do another fix

I strongly recommend lockless design.

ibigbug:
it's more change than I expected. I'll do another PR instead

ibigbug:
> I strongly recommend lockless design.

lock free is good and ideal, but some times make the code pattern more complicated than a simple mutex

xmh0511:
> > I strongly recommend lockless design.
> 
> lock free is good and ideal, but some times make the code pattern more complicated than a simple mutex

I don't appreciate that we implement a lock-free algorithm by ourselves. It's hard to implement correctly without doing a formal reasoning; that would bring more irreproducible bugs into the project. Moreover, the health-check function generally cannot have a high contention; implementing it with a correct lock-free algorithm has a low benefit.

However, I encounger we abstract them with mpsc, which is simpler.

Itsusinn:
Lockless programming is not complex or difficult to understand, nor does it involve algorithm-level concerns. CAS (Compare-And-Swap) is the main way to implement it.

Here is a related crate: https://crates.io/crates/arc-swap

If it doesn't meet your requirements, you can also try https://crates.io/crates/moka

Moka is a high-performance concurrent cache library built on top of a lock-free concurrent hash table (powered by CAS operations and epoch-based memory reclamation). Its core storage, cht::SegmentedHashMap, allows multiple threads to read and write simultaneously without blocking.

xmh0511:
> Lockless programming is not complex or complicated, nor is it algorithm-related. CAS (Compare-And-Swap) is the main way to implement it.
> 
> Here are https://crates.io/crates/arc-swap
> 
> If it doesn't fit requirement, https://crates.io/crates/moka could also work

arc-swap has some implementation bugs for guaranteeing correctness; see my post [issue](https://github.com/vorner/arc-swap/issues/200), not completely fixed.

> CAS (Compare-And-Swap) is the main way to implement it.

Besides that, you should ensure that all (non-atomic)operations are correctly ordered, which requires formal proof.

Itsusinn:
> arc-swap has some implementation bugs for guaranteeing correctness; see my post [issue](https://github.com/vorner/arc-swap/issues/200), not completely fixed.

 I gotta say it is less dangerous than deadlock

> > CAS (Compare-And-Swap) is the main way to implement it.
> 
> Besides that, you should ensure that all (non-atomic)operations are correctly ordered, which requires formal proof.

Thats based on how strict the scenario is. Clash-rs is read-most write-few. Relax ordering is enough

xmh0511:
> I gotta say it is less dangerous than deadlock

No, these bugs will result in use-after-free, which is not memory safe and is more harmful than a deadlock. 


> Thats based on how strict the scenario is. Clash-rs is read-most write-few. Relax ordering is enough

Relaxed ordering could lead to out-of-order execution; this should be analyzed in light of what clash-rs actually want to do.  I'm not familiar with this part of clash-rs. However, according to your referred crates and the current implementation, I suppose you need RCU, which does need ordering. In other words, the resource's drop and the use should be in order.

ibigbug:
@copilot resolve the merge conflicts in this pull request

Itsusinn:
> No, these bugs will result in use-after-free, which is not memory safe and is more harmful than a deadlock.

We do cross testings on most platforms, use after free will cause an UB during test?
Besides we should only use arcswap on basic types(or plain data types)

> I'm not familiar with this part of clash-rs. However, according to your referred crates and the current implementation, I suppose you need RCU, which does need ordering. In other words, the resource's drop and the use should be in order.

We are not talking about basic types （ int, uint or float）but abstract types （handler, manager） Copy is not suitable

Alas,  talking is cheap, show the code is more direct. I will try draft a proof-of-concept PR

xmh0511:
> We do cross testings on most platforms, use after free will cause an UB during test?

Yes, that's UB. However, UB means every result is possible. For example, it could cause segment fault, or it can behave as expected behavior you want, but it's not guaranteed.

ibigbug (approved):

PR #1423, opened 2026-05-17
config: tolerate multiple YAML shapes for reality.short-id

grill-glitch:
Make reality.short-id deserialization accept string / null / empty sequence / missing and normalize to empty string.

## Problem
Config parser currently expects `reality.short-id` to be a string. If users write `short-id:` (empty key) or an empty YAML sequence, serde errors with "invalid type: sequence, expected a string".

## Fix
Add a custom serde deserializer that accepts string / null / unit / empty sequence and normalizes empty cases to `""` (empty string). Existing non-empty short-id behavior unchanged.

## Rationale
Improves UX and tolerates common YAML styles; downstream `hex::decode("")` already returns empty `Vec<u8>`, so behavior is consistent.

## Changes
- `clash-lib/src/config/internal/proxy.rs`: added `deserialize_short_id` custom deserializer and annotated `RealityOpt.short_id` with `#[serde(deserialize_with = "deserialize_short_id", default)]`

## Testing
- `clash-lib` crate compiles successfully with the change
- Existing tests for short-id-too-long still apply
- Full test suite could not be run due to upstream dependency issues (forked rustls/boringtun API incompatibilities), unrelated to this change

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration parsing for the reality short-id setting: accepts plain strings, treats null or an empty sequence as an empty value, accepts single-item sequences as the value, and rejects multi-item sequences. Missing or default entries are now normalized to an empty value for consistent behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Watfaq/clash-rs/pull/1423?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/clash-lib/src/config/internal/proxy.rs
+++ b/clash-lib/src/config/internal/proxy.rs
@@ -1,5 +1,5 @@
 use crate::{Error, common::utils::default_bool_true, config::utils};
-use serde::{Deserialize, de::value::MapDeserializer};
+use serde::{de::value::MapDeserializer, Deserialize, Deserializer};
 use serde_yaml::Value;
 #[cfg(feature = "shadowquic")]
 use shadowquic::config::CongestionControl as SQCongestionControl;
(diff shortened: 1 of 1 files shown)


ibigbug:
doesn't making the field default simply work?

grill-glitch:
@ibigbug ,
if the config is
```yaml
short-id:
```
it will return:
`Error: invalid config: could not parse config content: invalid type: sequence, expected a string`

PR #1475, opened 2026-07-18
feat(tun): add FreeBSD tun support via FIB multi-table routing

lakako:
## Summary

Adds tun-mode support on FreeBSD (verified on FreeBSD 15.1) — previously clash-rs had no tun implementation for the `bsd` feature set (`clash-bin`'s `bsd` feature did not even enable `tun`), and a plain `tun: { enable: true }` config failed at startup with:

```
ERROR clash-lib/src/proxy/tun/runner.rs:283: tun initialization I/O error: No such file or directory (os error 2)
```

This PR closes that gap and makes tun fully usable as a FreeBSD L3 bypass-gateway (翻墙).

## Background: why FreeBSD needs a different approach

- FreeBSD has **no `SO_BINDTODEVICE`** (or usable `IP_BOUND_IF`), so the existing `platform/unix.rs` `must_bind_socket_on_interface` returns `unsupported platform: <iface>` on FreeBSD — the first outbound socket clash tries to bind fails.
- FreeBSD has **no Linux-style policy routing** (`ip rule` + `fwmark`). The natural FreeBSD equivalent of "capture all traffic into the tun while letting the proxy itself escape via the physical gateway" is **FIBs (multiple routing tables)** + `setfib(2)`:
  - **FIB 0** (the default routing table every unprivileged process uses) gets its default route moved onto the tun interface, so unrelated host traffic is captured by clash.
  - A separate **bypass FIB** (configurable via `route_table`, default `2468`) is populated with a copy of the host's *original* default gateway route. clash's whole process is moved onto this bypass FIB via `setfib(2)`, so every outbound socket it opens thereafter is routed by the physical gateway directly — no tun loop.
  - `setfib(2)` is process-wide (applies to all subsequent sockets, regardless of thread), so a single call after tun setup is enough; no per-connector plumbing is needed.

### `clash-lib/src/proxy/tun/routes/freebsd.rs` (new, ~620 lines)
Mirrors Linux's policy routing using FreeBSD FIBs:
- **`setup_policy_routing` (route_all=true):** move FIB 0 default → tun; install original gateway (with an on-link `-host` route via the original interface) into the bypass FIB; `setfib()` clash into the bypass FIB.
- **`setup_partial_fib` (route_all=false, per-listed `routes`):** leave FIB 0 default untouched; only push clash into the bypass FIB so DIRECT outbounds don't loop for the captured destinations (FreeBSD has no fwmark, so FIB is used instead of Linux's `not fwmark` rule).
- **`ensure_tun_device(name):** the fix for the ENOENT above. `tun-rs`'s FreeBSD backend opens `/dev/<tunN>` directly, but that device node does **not** exist statically on FreeBSD — devfs materialises it only while the matching interface exists. Pre-create the interface with `ifconfig <name> create` before tun-rs opens the node.
- **`restore_stale_state()`:** self-heal a leaked routing state at startup. `TunRunner::shutdown()` is only invoked via the external controller/IPC — `SIGTERM` / `SIGKILL` do **not** trigger it, so a killed clash leaves FIB 0's default dangling on a (now-destroyed) tun and the bypass FIB dirty. This bricked host networking and made the next start abort with "requires an existing IPv4 default route" (because `route -n get default` then failed). On startup, if a state file is present we restore FIB 0's default from the saved gateway, drop the bypass-FIB routes, remove the state, and proceed with a fresh setup — **no manual recovery needed after a crash / kill -9**.
- **`run_idempotent()`:** tolerate "File exists" / "already in table" on `route add` so re-installing after a dirty routing table doesn't fail.
- **State persistence** (`/tmp/clash-rs-fib-tun.state`): saves the original v4/v6 gateway + the `route_all` flag so clean-up and self-heal restore correctly even across crashes.
- **Failure rollback:** if installing the bypass default fails mid-setup, FIB 0's default is rolled back so the host never loses connectivity mid-setup.
- Raises `net.fibs` at runtime if needed (only increases are allowed; max 65536).

### `clash-lib/src/proxy/tun/routes/mod.rs`
Dispatch to the new `freebsd` module; exclude freebsd from the no-op `other` cfg; re-export `ensure_tun_device`.

### `clash-lib/src/proxy/utils/platform/unix.rs`
`must_bind_socket_on_interface` is a **no-op** on FreeBSD. There is no `SO_BINDTODEVICE` on FreeBSD, and loop avoidance is already handled by the process-level `setfib(2)` — binding to the physical interface here would only loop clash's outbounds back into the tun.

### `clash-lib/src/proxy/tun/runner.rs`
In the new-tun branch, call `routes::ensure_tun_device(&tun_name)` on FreeBSD just before `DeviceBuilder::build_async()` when the tun does not already exist.

### `clash-bin/Cargo.toml`
Enable the `tun` feature in the `bsd` feature set (previously `bsd = ["shadowsocks", "tuic", "ring", "clash-lib/zero_copy"]` — tun was never compiled).

## Verification (FreeBSD 15.1)

Build: `cargo build --release --no-default-features --features bsd -p clash-rs` ✅

End-to-end with a real subscription config (`route-all: true`, `device-id: dev://tun0`, trojan proxies, `MATCH,Final`):
- **Fresh host** (no `/dev/tun0`, no state): clash creates tun0, sets up FIB routing, `http(s)://www.google.com/generate_204` from the host → **204** via the trojan proxy (traffic: host → FIB0 default → tun0 → clash netstack → rule FINAL → trojan → physical gateway via bypass FIB). ✅
- **`kill -9` crash simulation:** host loses FIB0 default → `ping 1.1.1.1` from a *non-clash* terminal → "No route to host".
- **Restart clash with no manual cleanup:** logs `found leftover state ... restoring` → `self-heal: restored FIB 0 IPv4 default -> 192.168.1.1` → `cleaned stale state` → `created tun interface tun0` → fresh tun setup → `google.com/generate_204` → **204** again. ✅
- Host network is preserved across all stages (SSH never dropped).

A `partial` mode (`route-all: false`, `routes: [1.1.1.1/32]`) was also verified: only the listed destination is captured; clash's own DIRECT outbounds escape via the bypass FIB; host FIB0 default is untouched.

## Notes / limitations

- `setfib(2)` is process-global, so this scheme (like `route_all`) inherently anchors clash's *own* outbound traffic to a single bypass FIB. That matches the Linux `fwmark` story. Per-socket `SO_SETFIB` plumbing would be a much larger change and isn't needed for the standard tun use case.
- ICMP via tun does not work (clash-netstack/smoltcp doesn't forward ICMP) — same as on Linux. TCP/UDP 翻墙 works.
- Requires root (or `CAP_NET_ADMIN` equivalent) to create tun interfaces and alter routing tables — same as on Linux.

## Checklist

- [x] Builds with `--features bsd` (and with the default `bsd` feature set in `clash-bin`)
- [x] Doesn't change behaviour on Linux / macOS / Windows (all changes are `#[cfg(target_os = "freebsd")]` except the `bsd` feature flag addition)
- [x] Crash-safe: kill -9 leaves the host bricked but the next start self-heals; graceful shutdown via controller (which invokes `shutdown()`/`maybe_routes_clean_up`) is unchanged
- [x] State file written under `$TMPDIR` / `/tmp` so it survives process crashes

## Summary by CodeRabbit

* **New Features**
  * Added FreeBSD support for TUN-based routing using system routing tables.
  * Automatically creates missing TUN devices on FreeBSD before startup.
  * Introduced persistent routing state to recover cleanly after abrupt shutdowns.
  * Improved full vs partial routing behavior, including bypass handling.
* **Bug Fixes**
  * Prevented unsupported-platform errors when binding outbound sockets on FreeBSD.
  * Enhanced cleanup to restore the original network default routes (when full routing is enabled).
* **Chores / Build**
  * Updated the FreeBSD build target matrix by removing the 32-bit i686-unknown-freebsd entry.


Itsusinn:
Did you test it on a real BSD machine or a visual machine？

lakako:
> Did you test it on a real BSD machine or a visual machine？

Only test on FreeBSD 15.1 x64

image

ibigbug (approved):
lgtm - can you plz address the build failures, thanks!

PR #1477, opened 2026-07-21
refactor(dispatcher,transport): dedup tracked streams + TransportLayer enum

ibigbug:
## Summary

Behavior-preserving simplification of the proxy stream / transport layers — fewer traits, one stream type, capabilities as methods, no `downcast_rs` / `Box<dyn Transport>`.

**1. Dedup `TrackedStream`/`TrackedDatagram` boilerplate** — single `account_*` helper (6→1), single `poll_closed()` (7→1), `ReadTracker`/`WriteTracker`→one `DirCopy`.

**2. `Box<dyn Transport>` → closed `TransportLayer` enum** — static dispatch, no per-layer heap alloc, same runtime composition. All handlers + converters. Reality XTLS splice preserved.

**3. Shadowsocks plugins through `TransportLayer`, delete `Sip003Plugin`** — the obfs/shadow-tls/v2ray plugins were the last boxed transport (via a redundant one-method `Sip003Plugin` blanket-over-`Transport` trait); routed through the enum variants and the trait + module deleted.

**4. `ClientStream` downcast → `underlying_socket()` capability** — dropped `downcast_rs` on the splice path for a defaulted capability method.

**5. Collapse `TrackedStream` into `ChainedStreamWrapper`** — two-layer wrapper → one, with `tracking: Option<StreamTracking>` installed once by the dispatcher. Untrack fires exactly once (outer box only). `TrackedStream`/`TrackedDatagram` deleted.

**6. Merge `ClientStream` into `ProxyStream`** — one stream trait. `ProxyStream` is now non-blanket (carries `underlying_socket()`; only `TcpStream` overrides), with `Send`/`Sync` expressed at the boxed sites (`AnyStream = Box<dyn ProxyStream + Sync>`, `ChainedStream: ProxyStream + Sync`) rather than as trait bounds — so inbound streams that are `Send` but not `Sync` (`TokioIo<Upgraded>`) still qualify. ~25 stream types get an explicit (mostly empty) `impl ProxyStream`; the generic `ChainedStreamWrapper<T>` impl covers the bulk.

## Test plan
- [x] `cargo build -p clash-lib` — all-features / default / `--features zero_copy` clean
- [x] `cargo clippy -p clash-lib --all-features` — no new warnings
- [x] `cargo test -p clash-lib --lib --all-features` — 244 passed; 1 pre-existing failure (`tailscale_live_auth_key`, needs live creds)

## Deferred
`async_trait` removal below the dyn seams (native AFIT) — a separate, larger migration.


ibigbug:
Thanks for the review. Addressed the findings:

- **Download overcount (tracked.rs poll_read):** now counts only the delta `buf.filled().len() - before` and only on `Poll::Ready(Ok(()))`, so pre-existing buffer bytes and Pending/error polls are no longer counted. (30d666ad)
- **Waker registration for close detection:** `poll_closed` now takes `cx` and polls the oneshot via `Pin::new(&mut close_notify).poll(cx)` instead of `try_recv()`, so idle connections wake when `Manager::close` fires. Applied to the stream `poll_read/write/flush/shutdown` and datagram `poll_next/poll_ready/poll_flush/poll_close`. `start_send` has no task context, so it keeps a plain `try_recv`. (30d666ad)
- **Skip accounting when n==0:** applied to `account_download`/`account_upload`. (earlier commit)

Note these three were pre-existing on `master` (this PR's refactor preserved the prior behavior), but they're real latent issues so fixed here. All builds (`--all-features`, default, `zero_copy`, `-F plus --cfg docker_test`), clippy, and the test suite pass.

PR #1480, opened 2026-07-22
fix(ws): percent-encode spaces in WebSocket paths

cedar-otter-47:
### What does this PR do?

When a WebSocket path contains spaces, the current implementation passes the unencoded path directly to the HTTP URI builder and panics with `InvalidUriChar` while constructing the request.

This PR:

- percent-encodes spaces as `%20` while preserving literal `.` and `..` path segments and the query string;
- converts invalid URI, header, and other request-construction errors into `InvalidInput` instead of panicking the proxy task;
- adds regression coverage using an in-memory WebSocket handshake, plus malformed URI and header cases.

### Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring / cleanup
- [ ] Other

### Checklist

### Validation

The same three regression tests cover space encoding and verify that malformed URIs and headers no longer panic the task:

- 6249eb5372a24b7273a1e68fe07a62a7be27edd5: [failing run](https://github.com/cedar-otter-47/clash-rs/actions/runs/29944303765/job/89005682215) — before the fix, 0 passed and 3 failed on the targeted behaviors.
- 49d109dfa2305ead6768030c373da8a5df931aef: [passing run](https://github.com/cedar-otter-47/clash-rs/actions/runs/29944554853/job/89006525438) — with the fix and the same tests, 3 passed and 0 failed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket URL construction by percent-encoding paths more safely (including spaces, non-ASCII characters, and preserving already-encoded sequences).
  * Invalid WebSocket inputs (e.g., bad host or malformed header names) now return clear input errors instead of panicking.
  * WebSocket connection setup/request issues are now properly reported to callers.
* **Tests**
  * Added Tokio tests to verify correct handshake request target encoding and confirm invalid inputs fail gracefully.


Itsusinn:
There must be a common crate which can do usual url escape

cedar-otter-47:
> There must be a common crate which can do usual url escape

Thanks. The initial implementation used the `url` crate, but parsing the URL normalizes literal `.` and `..` path segments, changing the configured WebSocket path.

I checked other Clash implementations, and their URL handling appears to preserve these segments. After looking for a Rust alternative with the same behavior, I switched this to `percent-encoding`, which performs standard percent-encoding without parsing or normalizing the path.

ibigbug (approved):
lgtm

PR #1482, opened 2026-07-22
fix(hysteria2): refresh the UDP poller after port hops

cedar-otter-47:
### What does this PR do?

After a Hysteria2 port hop, the send path switches to a new UDP socket, but Quinn's writable poller remains bound to the old socket. If the new socket returns `WouldBlock`, the old poller may continue reporting readiness and cause the send task to busy-loop.

This PR makes the writable poller follow the current socket and clones `Transmit` before changing its destination port, avoiding mutation through a shared reference. The regression test verifies that only the new socket is polled after a socket switch.

### Type

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring / cleanup
- [ ] Other

### Checklist

### Validation

The targeted regression test at dae7b0a3719c0b8a775c2a68757435517d58151f [shows the stale poller being used](https://github.com/cedar-otter-47/clash-rs/actions/runs/29946498740/job/89013089315): the old socket was polled once and the new socket was not polled.

With the same test unchanged, 43ea7a036dfed84f787d9aa6ebe82cc5c5dd9365 [shows the poller following the new socket](https://github.com/cedar-otter-47/clash-rs/actions/runs/29946826594/job/89014214492): 1 passed and 0 failed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UDP connection handling when switching between hop connections.
  * Ensured write-readiness checks use the currently active connection.
  * Improved reliability when sending UDP packets after a connection change.
* **Tests**
  * Added coverage for polling behavior during UDP hop switches.


ibigbug (approved):
thanks!

PR #1491, opened 2026-07-31
feat(tun): add system (kernel) TCP stack option

Ember-Moth:
### What does this PR do?

Adds a `stack` option to the TUN inbound, offering a system-stack backend alongside the existing smoltcp userspace stack:

```yaml
tun:
  enable: true
  stack: system   # default: smoltcp (alias: gvisor); system also accepts
                  # "mixed" for clash-meta config compatibility
```

**How it works**: TCP packets read from the TUN device are source-NATed in
place (RFC 1624 incremental checksum update) and redirected to a kernel TCP
listener bound on the TUN gateway address, so the OS owns all TCP state —
congestion control, window scaling, retransmission. Accepted streams are
dispatched through the normal `Dispatcher::dispatch_stream` path with the
original five-tuple recovered from a sharded NAT table (this works out of
the box since `tokio::net::TcpStream` already implements `ProxyStream`).

**What deliberately stays on the userspace path**: UDP packets are forwarded
through the existing `watfaq_netstack::UdpSocket` packet codec and
`handle_inbound_datagram`, so the datagram path — including DNS hijack — is
byte-for-byte identical to the smoltcp stack. (In clash-meta terms this
backend is closest to `mixed`, hence the alias.)

### Measured results

Linux 6.18 (WSL2), i7-12650H, release build **without** `telemetry`, two
isolated network namespaces joined by veth, iperf3, TUN `txqueuelen` 10000
for both stacks, median of per-second steady-state intervals. Baseline
(direct veth, no TUN) was ~66 Gbps.

| streams | smoltcp | system | delta | smoltcp RSS | system RSS |
|--------:|--------:|-------:|------:|------------:|-----------:|
| 1  | 3.39 Gbps | 4.82 Gbps | **+42.2%** | 18 MB  | 18 MB |
| 16 | 4.36 Gbps | 4.68 Gbps | **+7.3%**  | 47 MB  | 18 MB |
| 64 | 4.13 Gbps | 4.09 Gbps | −0.8%      | 137 MB | 19 MB |

Memory is the more dramatic difference and it is structural: the system
stack allocates no per-connection buffers, because the kernel owns TCP
state — a connection costs one ~100-byte NAT entry. The smoltcp stack
allocates **1 MiB per TUN connection** (2 × 256 KiB smoltcp `SocketBuffer`
at `clash-netstack/src/tcp_listener.rs:259` plus 2 × 256 KiB
`LockFreeRingBuffer` at `:65`, all zero-initialised), which is the 137 MB
above and keeps scaling linearly.

### Getting there required batching the device I/O

Worth recording, because the first cut was *slower* than smoltcp
(−6.6% / −40.5% / −43.6% at the same three concurrency levels) and the
reason is specific to this design.

The system stack has to write every packet **back** into the TUN device so
the kernel can deliver it to the listener: uplink is rewritten and
re-injected, and the listener's downlink reply is routed out the TUN, read,
rewritten, and re-injected. The smoltcp stack never re-injects uplink — its
userspace TCP consumes it directly. For the same payload the system stack
therefore issues roughly **2× the device operations**, so with one syscall
per packet it saturated the device queue under concurrency and paid for the
drops in TCP retransmissions.

Fixed with `tun_rs`'s batched offload API: `recv_multiple` takes one GRO
read (up to 64 KiB) and splits it into individual packets, `send_multiple`
coalesces outgoing packets back into fewer, larger writes.

Two things are worth knowing if you touch this:

- **`writev` cannot batch here.** On a TUN character device a vectored write
  still assembles exactly one packet, so GRO/GSO is the only way to amortise
  the syscall.
- **Offload is enabled only for `stack: system`, and only on Linux.** It
  prepends a virtio header to every device read and write, which the
  userspace stack's framed codec does not expect. `recv_multiple` /
  `send_multiple` also degrade to single-packet I/O when the device has no
  virtio header, so the batched path stays correct for `fd://` devices and
  reused interfaces where the builder never set the flag.

I also implemented and measured a ping-pong arrangement (reader and writer
in separate tasks handing buffer sets back and forth, so the reader never
stalls on a write). It was clearly worse — **+7.6% / −19.0% / −32.0%** — so
it is not in this PR: the per-batch channel round-trip costs more than the
write stall it removes. Left a note in the code so nobody re-derives it.

### Known remaining rough edge

At 64 streams the system stack still records ~74 k device TX drops while
matching smoltcp's throughput. Raising `txqueuelen` past the default absorbs
them (drops reach zero and the system stack goes slightly ahead), so it is a
burst-absorption limit rather than a throughput ceiling. Note the default
`txqueuelen` of 500 is too small for multi-gigabit for **both** stacks —
smoltcp drops ~39 k/67 k packets at 16/64 streams there too.

### Behavioral differences vs the smoltcp stack

- `gateway` must match the actual TUN interface address; the adjacent
  address (`gateway + 1`) is reserved as the NAT source and must be free.
  The same applies to `gateway_v6` if set. A reused device whose address
  does not match is now rejected with a clear config error.
- ICMP echo is answered only for the gateway address itself — pinging
  arbitrary remote IPs through the TUN gets no reply
- fragmented IP packets are dropped (same as leaf)
- IPv6 UDP packets carrying extension headers are dropped, because the
  shared UDP codec assumes a fixed 40-byte IPv6 header

### Type

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring / cleanup
- [ ] Other

### Checklist

- [x] Tests added or not needed — 12 unit tests covering the IPv4/IPv6 TCP
      rewrite (including incremental-checksum correctness against a full
      recomputation), the downlink restore branch and an unmapped downlink
      packet, the NAT table and its cleanup race, ICMP/ICMPv6 replies, NAT
      address derivation, and config parsing plus conversion for every alias
- [x] Docs updated or not needed — the new field is doc-commented in
      `def.rs`, so it lands in the generated config docs

### Testing

- Full `clash-lib` suite passes (267 tests); `cargo clippy --all
  --all-features` clean
- Functional end-to-end on Linux: five-tuple correctly restored
  (`[TCP] 198.19.0.1:58811 -> 1.1.1.1:80 to DIRECT`), HTTP and HTTPS both
  succeed, concurrent connections fine, DNS round-trips through the UDP
  codec path, TUN device and routes cleaned up on shutdown
- Throughput as tabulated above

**Not yet tested on Windows (wintun) or macOS** — both take the
non-batched per-packet path, which is the code this PR started from.

### Three benchmarking issues found along the way

Unrelated to this feature, but they cost me real time and will bite the next
person. Happy to send separate PRs.

1. **`--all-features` makes throughput numbers meaningless.** It enables
   `telemetry`, and the resulting OTLP export attempts cost roughly **500×**
   throughput in my runs — 5.5 Mbps versus 2.9 Gbps for the same config,
   with CPU pinned across multiple cores and RSS climbing steadily. Both
   `benchmark.yml` and `proxy-throughput.yml` build with `--all-features`.
2. **The CI TUN benchmark cannot see this feature.**
   `bench/tun-benchmark.yaml` has no `stack:` field, so it only ever
   exercises smoltcp; the run on this PR reported 1.28 Gbps, which is the
   unchanged userspace path. A `system` variant would be needed for coverage.
3. **`bench/run_tun_benchmark.py` fails silently on hosts that rewrite
   packet marks.** It relies on fwmark policy routing, and on WSL2 the
   `WSLOUTPUT` nftables chain sets `meta mark 0x1` on every output packet,
   clobbering `so-mark: 666` and routing clash-rs's own egress back into the
   TUN. The only symptom is `iperf3 test timed out`. Running the whole
   benchmark inside a network namespace avoids it, since nftables rulesets
   are per-netns.

Also: the global `interface` config option is documented as "not implemented
yet" in `def.rs:486` and silently does nothing, which is easy to trip over
when trying to pin an outbound interface for a test.

## Summary by CodeRabbit

* **New Features**
  * Added configurable TUN networking stacks, with userspace mode enabled by default and optional system/kernel mode.
  * System mode supports IPv4 and IPv6 traffic, TCP and UDP forwarding, ICMP handling, and address translation.
* **Compatibility**
  * Legacy YAML stack names remain supported through configuration aliases.
* **Bug Fixes**
  * Improved validation for TUN devices and malformed or unsupported packets.
  * Improved connection cleanup, packet handling, and cancellation behavior in system-stack mode.

### DIFF
--- a/clash-bin/tests/data/config/full.yaml
+++ b/clash-bin/tests/data/config/full.yaml
@@ -230,6 +230,12 @@ tun:
 
   # mtu: 1500                       # MTU for the TUN interface
 
+  # TCP/IP stack: "smoltcp" (default, alias "gvisor") is the userspace stack;
+  # "system" (alias "mixed") hands TCP to the kernel — faster and far lighter
+  # on memory at high connection counts, but requires `gateway` to match the
+  # real interface address and only answers ICMP echo to the gateway itself.
+  stack: smoltcp
+
   so-mark: 0                        # fwmark for TUN-originated packets (Linux only)
   route-table: 2468                 # Policy routing table (Linux only)
 
--- a/clash-lib/src/config/def.rs
+++ b/clash-lib/src/config/def.rs
@@ -52,19 +52,21 @@ impl Default for DnsHijack {
 pub enum TunStack {
     /// Userspace network stack based on smoltcp
     #[default]
-    #[serde(alias = "gvisor", alias = "Smoltcp")]
+    #[serde(alias = "gvisor", alias = "gVisor", alias = "Smoltcp")]
     Smoltcp,
     /// Kernel TCP stack: TCP flows are source-NATed in place to a local
     /// listener so the OS owns TCP state (congestion control, window
-    /// scaling), which usually outperforms the userspace stack. UDP and
-    /// ICMP stay on the userspace path.
+    /// scaling), which usually outperforms the userspace stack. UDP stays on
+    /// the userspace path.
     /// # Note
     /// - requires `gateway` to match the actual TUN interface address
     /// - the address right after `gateway` within the prefix is reserved as the
     ///   NAT source address and must not be used by anything else
     /// - if `gateway_v6` is set, the address right after it within its prefix
     ///   is reserved the same way
-    #[serde(alias = "mixed", alias = "System")]
+    /// - only ICMP echo to the `gateway` address is answered; pings to other
+    ///   addresses through the TUN get no reply, unlike the `smoltcp` stack
+    #[serde(alias = "mixed", alias = "Mixed", alias = "System")]
     System,
 }
 
--- a/clash-lib/src/config/def.rs
+++ b/clash-lib/src/config/def.rs
@@ -46,6 +46,26 @@ impl Default for DnsHijack {
     }
 }
 
+/// The TCP/IP stack implementation used by the TUN inbound
+#[derive(Serialize, Deserialize, Default, Copy, Clone, PartialEq, Eq, Debug)]
+#[serde(rename_all = "lowercase")]
+pub enum TunStack {
+    /// Userspace network stack based on smoltcp
+    #[default]
+    #[serde(alias = "gvisor", alias = "Smoltcp")]
+    Smoltcp,
+    /// Kernel TCP stack: TCP flows are source-NATed in place to a local
+    /// listener so the OS owns TCP state (congestion control, window
+    /// scaling), which usually outperforms the userspace stack. UDP and
+    /// ICMP stay on the userspace path.
+    /// # Note
+    /// - requires `gateway` to match the actual TUN interface address
+    /// - the address right after `gateway` within the prefix is reserved as the
+    ///   NAT source address and must not be used by anything else
+    #[serde(alias = "mixed", alias = "System")]
+    System,
+}
+
 #[derive(Serialize, Deserialize, Default)]
 #[serde(rename_all = "kebab-case")]
 pub struct TunConfig {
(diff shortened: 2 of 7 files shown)


ibigbug:
@copilot review

PR #1503, opened 2026-08-22
feat(tls): apply client-fingerprint instead of warning about it

Rushan4eg:
> **Depends on [Watfaq/rustls#11](https://github.com/Watfaq/rustls/pull/11)** — the mechanism lives there, the browser table lives here.
>
> Draft until that merges. So the branch can actually build and be reviewed, `Cargo.toml` points at that PR's branch on my fork, marked in place as not for merge; it goes back to `Watfaq/rustls` the moment #11 lands.

## Why

Every VLESS or AnyTLS proxy with `client-fingerprint` set gets this:

```
WARN clash-lib/src/proxy/converters/vless.rs:35:
client-fingerprint (uTLS) is not yet implemented, ignored for main-out
```

Reality connects and carries traffic, but the ClientHello is recognisably rustls: no GREASE, no `signed_certificate_timestamp`, no ALPS, and a cipher list that is exactly the fifteen-minus-six suites rustls happens to implement. That separates it from a browser before a byte of application data moves, which is the one thing Reality exists to prevent.

## What changed

`client-fingerprint` is now applied, for VLESS (Reality and plain TLS) and for AnyTLS.

The numbers come from [uTLS](https://github.com/refraction-networking/utls) `HelloChrome_133`, not from memory — `clash-lib/src/common/tls_fingerprint.rs` is the whole table, and it is the only file that knows anything about browsers.

| accepted value | profile |
|---|---|
| `chrome`, `edge`, `android`, `360`, `qq` | Chrome 133 |
| `random`, `randomized` | rustls' own hello, plus GREASE |
| anything else | falls back to `random` **with a warning** |

That last row is deliberate. Firefox and Safari are real profiles that are simply not modelled here; quietly substituting Chrome for them would be worse than saying so, because a config asking for Firefox and getting Chrome is a mismatch nobody would think to look for.

### What the Chrome profile sets

- **cipher suites** — Chrome's fifteen, in order. Six of them (static RSA and CBC) rustls cannot negotiate and will never be selected; they are advertised because the count and the order are the first half of the fingerprint.
- **GREASE** — cipher list, supported groups, key shares, supported versions, and two extensions.
- **supported groups** — X25519MLKEM768, X25519, P-256, P-384. rustls derives the key shares from that order on its own, and the pair it produces is the pair Chrome sends.
- **signature algorithms** — Chrome's eight, in order, via a verifier wrapper that also narrows what will be accepted. Same trade the browser makes.
- **`renegotiation_info`** — rustls signals "no renegotiation" with the SCSV in the cipher list; browsers use the extension. Pinning the suite list removes the SCSV, so the extension takes its place.
- **`signed_certificate_timestamp`** and **ALPS** (`0x44cd`, the codepoint Chrome 133 uses) as verbatim extensions.
- **ALPN** `h2, http/1.1` — only when the caller has none of its own. A transport that must negotiate h2 has a functional requirement, and looking like a browser does not outrank it.
- **GREASE ECH** — a placeholder full of random data, X25519, matching what browsers send. Off without the `aws-lc-rs` feature: rustls only ships HPKE for that provider.
- **`compress_certificate`** — brotli, the only algorithm Chrome offers. See the note on the `brotli` feature below.
- **no padding** — Chrome 131 dropped it; with GREASE ECH the hello no longer lands in the range padding existed for.

## Measured, not just asserted

The tests say the profile is assembled correctly. This says what actually left the socket, read back from `tls.peet.ws`:

| | JA4 |
|---|---|
| clash-rs today | `t13d1011h1_61a7ad8aa9b6_f9531d972513` |
| this branch, no `client-fingerprint` | `t13d1012h1_61a7ad8aa9b6_41631feb4e62` |
| this branch, `client-fingerprint: chrome` | `t13d1516h1_8daaf6152771_d8a2da3f94cd` |

The middle row is the whole cost of the `brotli` commit, stated below: one extension more on every connection, and the cipher hash unchanged.

`1516` is fifteen cipher suites and sixteen extensions, which are Chrome's counts. `h1` rather than `h2` is the probe's own doing — it speaks HTTP/1.1 so it can read the reply, and JA4 records the ALPN it saw.

The raw form is the part worth reading:

```
t13d1516h1_002f,0035,009c,009d,1301,1302,1303,c013,c014,c02b,c02c,c02f,c030,cca8,cca9
          _0005,000a,000b,000d,0012,0017,001b,0023,002b,002d,0033,44cd,fe0d,ff01
          _0403,0804,0401,0503,0805,0501,0806,0601
```

Cipher list, extension list and signature algorithms: field for field what uTLS records for `HelloChrome_133`.

## On hardware, and what it changed

Flashed onto a GL-XE300 (`mips_24kc`, OpenWrt 25.12.5) running podkop with a real VLESS+Reality outbound, and the ClientHello read back off the wire with tcpdump:

```
ClientHello, 299 bytes, session_id 32 bytes
cipher suites (16, 15 real):
  GREASE 1301 1302 1303 c02b c02f c02c c030 cca9 cca8 c013 c014 009c 009d 002f 0035
extensions: GREASE, session_ticket, psk_key_exchange_modes, key_share,
  extended_master_secret, compress_certificate, alpn, supported_versions,
  status_request, supported_groups, server_name, signature_algorithms,
  ec_point_formats, renegotiation_info, signed_certificate_timestamp,
  application_settings, GREASE
JA4 counts: 15 ciphers, 15 extensions
```

Tunnel healthy throughout — 148, 155, 157 ms across three delay tests — and the `client-fingerprint (uTLS) is not yet implemented` line is gone from the log.

**But the profile is provider-dependent, and routers get the reduced one.** The MIPS targets build with `-F ring`, and against x86 with aws-lc-rs the same code produced sixteen extensions and five groups. Two things are missing on ring, both for the same reason:

| | on ring | why |
|---|---|---|
| `encrypted_client_hello` | absent | rustls ships HPKE only for aws-lc-rs |
| X25519MLKEM768 | absent from groups and key shares | `ring` has no ML-KEM |

So on a router the hello is `15 ciphers / 15 extensions` against Chrome's `15 / 16`, and the group list is X25519, P-256, P-384 where Chrome leads with the post-quantum hybrid.

Neither is silent — the code says so:

```
WARN tls_fingerprint.rs: key exchange group X25519MLKEM768 is unavailable, dropped from the hello
WARN tls_fingerprint.rs: GREASE ECH needs the aws-lc-rs provider, which this build does not have
```

Worth knowing before anyone assumes a router build gets the same disguise a desktop build does. Closing either gap means HPKE and ML-KEM for `ring`, which is upstream rustls work, not this PR.

## The one line that touches everything else

```toml
rustls = { version = "0.23", default-features = false, features = ["brotli"] }
```

Chrome offers brotli certificate compression. Without a decompressor rustls omits the extension entirely, and that was the single extension standing between `1515` and `1516`.

It is a separate commit because it is a separate decision, and not a free one: it applies to every TLS client in clash-rs, not only fingerprinted ones. The default hello goes from eleven extensions to twelve, and a brotli decompressor is linked in. Both are arguably improvements — smaller handshakes, and a default hello marginally less unlike a browser's — but dropping the commit is a supported outcome. The cost of dropping it is one extension out of sixteen.

## No fingerprint, no change

Without `client-fingerprint` nothing about the hello moves, SCSV included. There is a test asserting it, because a feature like this quietly altering every other connection would be a bad trade.

`TlsClient::new` keeps its signature; the fingerprint goes through `new_with_fingerprint`, so trojan, vmess, socks5, websocket and the rest are untouched.

## Tests

Nineteen, and eleven of them read the bytes that actually left a `ClientConnection` rather than the struct that produced them — a profile correct in a struct and lost on the way to the socket would pass every other kind of test:

- Chrome's cipher list arrives in order, GREASE in front, sixteen entries
- GREASE reaches supported groups, supported versions and key shares
- three key share entries: GREASE, X25519MLKEM768, X25519 — what Chrome sends
- `signed_certificate_timestamp`, ALPS and `renegotiation_info` present with the right bodies
- the SCSV is gone and `renegotiation_info` replaced it
- ALPN is `h2, http/1.1`
- `compress_certificate` offers brotli and only brotli
- the GREASE ECH placeholder is there, with a 32 byte encapsulated key
- TLS 1.2 survived the ECH extension (rustls pins 1.3 for real ECH, and must not for GREASE)
- with no fingerprint: no GREASE anywhere, no browser extensions, SCSV still there
- an unknown profile name resolves to `random`, never to Chrome

```
cargo test -p clash-lib --lib --features aws-lc-rs,zero_copy,tuic
test result: ok. 19 passed; 0 failed
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -128,6 +128,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloc-no-stdlib"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc7bb162ec39d46ab1ca8c77bf72e890535becd1751bb45f64c597edb4c8c6b3"
+
+[[package]]
+name = "alloc-stdlib"
+version = "0.2.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0e76a019e91224d279006ff972f1e984179a6e9feb050adba6ce8274aef23195"
+dependencies = [
+ "alloc-no-stdlib",
+]
+
+[[package]]
 name = "alloca"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -250,7 +265,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -261,7 +276,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1096,6 +1111,27 @@ name = "bounded-integer"
 version = "0.5.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "102dbef1187b1893e6dfe05a774e79fd52265f49f214f6879c8ff49f52c8188b"
+
+[[package]]
+name = "brotli"
+version = "8.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5cc91aac060a7a1e25823bdccbfb6af1875b88f17c6daac97894eed8207166b3"
+dependencies = [
+ "alloc-no-stdlib",
+ "alloc-stdlib",
+ "brotli-decompressor",
+]
+
+[[package]]
+name = "brotli-decompressor"
+version = "5.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3a32acac15fe1967bc3986b2a6347dffc965602354ea6f450ad07e8bfd253583"
+dependencies = [
+ "alloc-no-stdlib",
+ "alloc-stdlib",
+]
 
 [[package]]
 name = "brutal-core"
@@ -2324,7 +2360,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1747ed5fb4ab3a9f8253da3401efe7ca8f8e5ec373b2e700ff55c3304d84e31f"
 dependencies = [
  "heck",
- "indexmap 1.9.3",
+ "indexmap 2.14.0",
  "itertools 0.14.0",
  "proc-macro-crate",
  "proc-macro2",
@@ -2488,7 +2524,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -2895,7 +2931,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


Rushan4eg:
Closing — keeping this in a fork for now rather than asking for review time. Thanks.

PR #1504, opened 2026-08-26
fix: 20+ stability fixes for router/low-memory deployments

yangw9182-del:
## Summary
A set of stability fixes validated on a 128MB OpenWrt router (DSG-AX3000/IPQ5018) running as a transparent proxy for 30+ days without crash.

### Crash / panic hardening
- `redir`/`tproxy`/`must_into_socket_addr`: replace `unwrap()` with error-tolerant handling — no more panic on malformed connections
- `network_listener`: all inbounds (http/socks/mixed/redir/tproxy/tunnel) keep listening after accept errors
- `dispatcher`: harden `expect()` calls

### Concurrency
- `statistics_manager`: fix lock held across await (deadlock)
- `HealthCheck`: add `Drop` + `stopped` flag to stop background task on hot-reload (prevents task leak)
- copy buffer yield — avoids CPU starvation at high throughput

### Stability
- `session.rs`: clamp domain to 255 bytes, harden match guard
- `urltest`: backoff + tolerance support + proper force-switch behavior
- shadowsocks/anytls outbound: timeout & decode fixes

### Memory
The memory-leak after speed test is fixed with configurable buffers.

## Test
- Device: DSG-AX3000 (IPQ5018, 128MB RAM, OpenWrt 19.07)
- Mode: transparent proxy (redir 7892 + tproxy 7893) + fake-ip DNS
- Result: continuous 30+ days, no crash, RSS stable at ~17MB

## Note
Additive env vars (`CLASH_RS_MEM_LIMIT_MB`, `CLASH_RS_HEALTHCHECK_CONCURRENCY`) are off by default, safe to merge.

## Summary by CodeRabbit

* **New Features**
  * Added configurable memory limits, pressure handling, DNS cache cleanup, and closed-flow limits.
  * Added adaptive URL-test proxy selection, traffic-aware switching, forced fastest-proxy selection, and manual locking.
  * Added configurable AnyTLS buffer sizes and improved ARMv7 release builds.
  * External dashboards now fall back to embedded dashboards when unavailable or empty.

* **Bug Fixes**
  * Improved listener resilience so transient connection errors no longer stop services.
  * Fixed DNS, fake-IP, datagram, domain-address, and transport edge cases that could cause failures or hangs.
  * Memory API now reports the configured limit accurately.


ibigbug:
thanks! this is huge and looks great, give me some time to look into it

ibigbug:
@copilot resolve the merge conflicts in this pull request

ibigbug:
@copilot resolve the merge conflicts in this pull request

ibigbug:
i'm going to split this into smaller PRs so it's easier to review, i'll try to preserve the raw commits so the history statys intact. if you are willing to, you can also do it yourself and make few more smaller PRs

ibigbug:
# Code Review: PR [#1504](https://github.com/Watfaq/clash-rs/pull/1504) — 20+ Stability Fixes

**Author:** `@yangw9182-del` | **54 files** | **+2,337 / −420** | **15 commits**

> [!IMPORTANT]
> This PR is too large for a single review pass. The maintainer (`@ibigbug`) has already indicated they plan to split it into smaller PRs. This review is structured to help prioritize which pieces are safe to land as-is and which need rework.

---

## Verdict Summary

| Category | Count | Severity |
|----------|-------|----------|
| 🔴 Must Fix (Blockers) | 7 | Critical / High |
| 🟠 Should Fix | 9 | Medium |
| 🟡 Nits / Style | 6 | Low |
| ✅ Good Changes | 10+ | — |

---

## 🔴 Must Fix (Blockers)

### 1. `.cargo/config.toml` — Chinese mirror registry + disabled certificate revocation

```toml
[source.crates-io]
replace-with = "rsproxy-sparse"

[http]
check-revoke = false
```

> [!CAUTION]
> **This must not be merged.** It replaces the crate registry for **all contributors** with a ByteDance mirror (`rsproxy.cn`) and disables Windows certificate revocation checks globally. This is a supply-chain security risk and a development environment leak. The author's local tooling config has bled into the PR.

**Action:** Remove the entire `.cargo/config.toml` diff.

---

### 2. CI Workflows — `permissions: contents: write` + unpinned actions

[build-armv7.yml](https://github.com/Watfaq/clash-rs/pull/1504/files#diff-build-armv7.yml), [build-memlimit.yml](https://github.com/Watfaq/clash-rs/pull/1504/files#diff-build-memlimit.yml)

> [!WARNING]
> - `permissions: contents: write` is over-privileged for a build-only workflow — these only upload artifacts, they don't create releases. Should be `contents: read`.
> - All GitHub Actions (`actions/checkout@v4`, `dtolnay/rust-toolchain@stable`, `goto-bus-stop/setup-zig@v2`, `actions/upload-artifact@v4`) are pinned to mutable tags, not commit SHAs. This is a CI supply-chain risk.
> - `build-memlimit.yml` has an unused `mem_limit_default` workflow input that is never passed to the build.
> - `cargo install cross || true` silently swallows installation failures.

**Action:** Pin actions to SHAs, downgrade permissions to `read`, wire or remove `mem_limit_default`, remove `|| true`.

---

### 3. `DEFAULT_BUFFER_SIZE` halved from 64KB → 8KB globally

```rust
// clash-lib/src/app/dispatcher/dispatcher_impl.rs
-const DEFAULT_BUFFER_SIZE: usize = 64 * 1024;
+const DEFAULT_BUFFER_SIZE: usize = 8 * 1024;
```

> [!CAUTION]
> This is an 8× throughput regression for **all** users, not just router deployments. 8KB forces AEAD-2022 ciphers (which frame at 16KB) into multiple encrypt/decrypt cycles per buffer fill, dramatically increasing CPU overhead. The original comment in the code explains this. The `experimental.tcp_buffer_size` config option already exists to let low-memory users opt in to smaller buffers.

**Action:** Revert to 64KB. Router users should set `experimental.tcp-buffer-size: 8192` in their config.

---

### 4. AnyTLS health check — raw TCP connect returns false-positive "alive"

```rust
// clash-lib/src/app/remote_content_manager/mod.rs
if matches!(outbound.proto(), OutboundType::Anytls) {
    // ...
    match connect {
        Ok(Ok(_stream)) => {
            return Ok((elapsed, elapsed)); // ← claims proxy is healthy
        }
```

> [!WARNING]
> A raw TCP connect to the server IP:port proves the IP is reachable, not that the proxy tunnel works. This marks a proxy as "alive" with a valid latency even when the AnyTLS handshake or the relay protocol is broken. The urltest group will then select this proxy and all traffic through it will fail.

**Action:** TCP connect should only be used as a *fallback latency estimate* when the full HTTP test explicitly succeeds through this proxy. If the AnyTLS handshake fails, the proxy must be marked dead.

---

### 5. Transparent proxy fallback in `url_test` bypasses the outbound under test

```rust
// URL test fallback: connect directly (no fw_mark) → transparent proxy catches it
let fallback_result = new_tcp_stream(addr, ..., None /* no fw_mark */)
```

> [!CAUTION]
> When `connect_stream` fails for a proxy, the code falls back to opening a raw TCP socket that gets captured by the transparent proxy (iptables REDIRECT). This means:
> - The health check result reflects the *transparent proxy's* latency, not the tested outbound's.
> - The outbound is marked "alive" even though its own tunnel is broken.
> - The hardcoded exclusion (`google.com`, `gstatic.com`) is fragile and doesn't cover other test URLs.
>
> This is the same class of bug as #4 — the health check lies about which proxy is working.

**Action:** Remove the transparent proxy fallback entirely. A failed `connect_stream` should result in a failed health check for that outbound.

---

### 6. `must_into_socket_addr` — silently returns `0.0.0.0:0` instead of panicking

```rust
// clash-lib/src/session.rs
self.try_into_socket_addr().unwrap_or_else(|| {
    tracing::warn!("...");
    SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), 0)
})
```

> [!WARNING]
> The function name `must_into_socket_addr` semantically promises a valid result or a crash. Returning `0.0.0.0:0` creates a silent data corruption path — callers will try to connect/send to `0.0.0.0:0`, which will either succeed (sending traffic into the void on localhost) or fail with a confusing error. The callers that used to panic should be individually audited: some should return `Err`, others should `continue`. A blanket fallback to `0.0.0.0:0` is worse than a crash because it's silent.

**Action:** At the two tproxy call sites, the PR already switches to `match try_into_socket_addr()`. Do the same for all other callers, then either remove `must_into_socket_addr` or keep the panic as a developer assertion.

---

### 7. `shadowsocks-rust` pinned revision change without explanation

```toml
-shadowsocks = { ..., rev = "c4d8d18527289d12470cad1ce718754c908a75fe" }
+shadowsocks = { ..., rev = "e543d9bc8caf88ae69411f452f10e3e37094bbbe" }
```

> [!IMPORTANT]
> The commit hash is bumped but the PR description doesn't explain why. The comment in the code says "SS2022 password 直接使用原始值" which implies a behavioral change in how passwords are handled. This needs:
> - A diff of what changed between the two revisions
> - Confirmation that existing SS2022 configs won't break
> - Whether this is related to a fix or just a version bump

---

## 🟠 Should Fix (Medium)

### 8. Memory pressure: `memory_usage()` called on every `track()` (per new connection)

```rust
pub async fn track(&self, item: Tracked, close_notify: Sender<()>) {
    let limit = self.mem_limit_bytes.load(Ordering::Relaxed);
    if limit > 0 {
        let current = self.memory_usage(); // ← syscall per connection
```

`memory_usage()` reads `/proc/self/statm` or equivalent on every new connection. Under load this can be thousands of calls/sec. Cache the value in an `AtomicU64` and refresh it in the `kick_off` ticker.

---

### 9. DNS cache size reduced from 4096 → 1024 without justification

```rust
-ResponseCache::new(4096, ...)
+ResponseCache::new(1024, ...)
```

This affects all users. 1024 entries is quite small for a DNS proxy serving 30+ devices. Should be configurable or left at the default.

---

### 10. `ClosedFlowInfo::from_tracker_info` clones many `String` fields

Every connection close now clones `host`, `network`, `source_ip`, `rule`, `rule_payload`, `country`, `asn`, and the entire proxy chain `Vec<String>`. For the stated goal of reducing memory, the PR should consider using `Arc<str>` or interning for the frequently-repeated strings (rule names, proxy chain entries).

---

### 11. `Mutex` (std, not tokio) used in `urltest::Handler` for `adaptive_state` and `manual_lock`

The url-test handler is called from async contexts. Using `std::sync::Mutex` in async code is fine if the critical section is very short and never awaits, which is the case here. However, poison propagation on panic is silently ignored with `if let Ok(...)` throughout, which means a panic in the scoring code would leave the adaptive state frozen forever. Consider using `parking_lot::Mutex` (no poisoning) or documenting the intentional poison-ignore.

---

### 12. `HealthCheck::Drop` only sets an `AtomicBool` — doesn't abort the task

```rust
impl Drop for HealthCheck {
    fn drop(&mut self) {
        self.stopped.store(true, Ordering::Relaxed);
    }
}
```

The background task checks `stopped` only on each tick interval (could be minutes). Between the drop and the next tick, the task continues running and may call `proxy_manager.check()` on stale proxy lists. Should also call `task_handle.abort()`.

---

### 13. `backoff_rounds` calculation uses subtraction that can underflow

```rust
if round - s.last_test_round < backoff {
```

If `last_test_round > round` (theoretically possible with concurrent checks or counter wrapping), this underflows on `u64`. Use `round.saturating_sub(s.last_test_round)`.

---

### 14. `url_test` manual lock stores a list index, not a proxy name

```rust
manual_lock: Mutex<Option<usize>>,
```

If the provider's proxy list changes (subscription refresh, provider update), the index may now point to a different proxy or be out of bounds. The `min(idx, proxies.len() - 1)` clamp silently locks to the wrong node. Store the proxy name instead and resolve to index at lookup time.

---

### 15. Chinese comments throughout new code

Large blocks of comments are written in Chinese (e.g., `// 先执行测速, 更新 proxy_manager 中的延迟数据`, `// 自适应 tolerance 配置常量`). The existing codebase uses English. Consistency matters for an open-source project with international contributors.

---

### 16. `socks/inbound/stream.rs` — `plen` read when `ulen == 0`

```rust
buf.resize(plen, 0);
s.read_exact(&mut buf[..]).await?;
```

When `ulen == 0`, `buf` was resized to 0. Then `plen` is read from `buf[0]` — but `buf` has length 0. The original code had `buf.resize(ulen, 0)` first. The fix uses `from_utf8_lossy` which is good, but the indexing logic for `plen` still reads from `buf[..1]` after the buf was resized to `ulen` bytes. If `ulen == 0`, `buf` is empty and `read_exact(&mut buf[..1])` would panic. This appears to be a pre-existing bug, but should be noted.

---

## 🟡 Nits / Style

### 17. `MemLimitMode` — manual `Default` impl instead of `#[derive(Default)]` + `#[default]`
The `None` variant can use the `#[default]` attribute (Rust 1.62+).

### 18. `pub_by_host` / `put_by_ip` — duplicated bidirectional-invariant code
The stale-mapping cleanup logic is identical in both methods. Extract into a shared helper.

### 19. `CopyBuffer::bytes_since_yield` uses `saturating_add` for a `u64`
This counter resets every 256KB. It can never overflow a `u64`. The `saturating_add` is unnecessary (harmless, but misleading).

### 20. `AnyTLS set_buffer_config` — no validation
`duplex = 0` or `relay = 0` would cause zero-size allocations.

### 21. Some docstrings removed from `urltest::Handler` impl methods
The `/// The name of the outbound handler` etc. comments were stripped. Minor, but inconsistent with the rest of the codebase.

### 22. `strategy_consistent_hashring` — retry loop simplification is correct
The change to a single `jump_hash` call is mathematically correct — `jump_hash` with `buckets > 0` always returns `[0, buckets)`.

---

## ✅ Good Changes (Worth Highlighting)

| Change | Why it's good |
|--------|---------------|
| **Listener restart loops** (all inbound handlers) | Critical fix — a single `accept()` error no longer kills the listener permanently |
| **`expect()` → `match` in `reverse_lookup` and `SocksAddr::Domain`** | Prevents panics on malformed connections |
| **`ClosedFlowInfo` lightweight struct** | Sound approach to fix the memory leak from retaining `Session`/`ProxyChain` in the ring buffer |
| **Deadlock fix in `statistics_manager`** (lock ordering) | Correct — the old code held `connections` lock across await points |
| **FakeIP bidirectional invariant enforcement** | Fixes real consistency bugs in the LRU cache |
| **FakeIP offset fix** (`offset: total - 1`, modulo `max - min + 1`) | Fixes off-by-one in IP allocation cycling |
| **Cooperative yield in `CopyBuffer`** | Excellent fix for CPU starvation on low-core devices |
| **gRPC `poll_read` waker fix** | Correct — returning `Pending` without registering a waker is a deadlock |
| **ShadowTLS `ReadState::WaitingHeader` transition** | Prevents silent stream corruption on non-APPLICATION_DATA records |
| **VLESS datagram reassembly** | Correct — TCP can deliver partial reads; the old code split UDP packets |
| **VMess datagram partial write loop** | Correct — the old code didn't handle partial writes |
| **Trojan datagram zero-byte write guard** | Prevents infinite spin loop |

---

## Recommendation

> [!IMPORTANT]
> **Do not merge as-is.** The `.cargo/config.toml` change alone is a blocker. The health check false-positive bugs (#4, #5) would cause traffic blackholes in production.
>
> Suggested split into smaller PRs:
> 1. **Crash hardening** (listener restarts, `expect()` removals, `must_into_socket_addr`) — safe to land
> 2. **Memory leak fix** (`ClosedFlowInfo`, `closed_flows_cap`) — safe to land
> 3. **Deadlock fix** (lock ordering in `statistics_manager`) — safe to land
> 4. **FakeIP fixes** (bidirectional invariant, offset) — safe to land, has test updates
> 5. **Transport fixes** (gRPC, ShadowTLS, VLESS, VMess, Trojan datagrams) — safe to land
> 6. **Cooperative yield** (`CopyBuffer`) — safe to land, low risk
> 7. **Memory pressure system** (`MemLimitMode`, `check_memory_pressure`) — needs rework (#3, #8, #9)
> 8. **Health check backoff + AnyTLS latency** — needs rework (#4, #5, #13)
> 9. **URL-test adaptive tolerance + manual lock** — needs rework (#14), Chinese comments (#15)
> 10. **CI workflows** — needs rework (#1, #2)

ibigbug:
Thanks for this great contribution! All the 20+ stability and performance improvements in this PR have been thoroughly reviewed, refined, and merged into \master\ via modular PRs (#1506, #1507, #1508, #1509, #1510, #1511, #1512, #1513, #1525). Closing this as fully resolved.

